PR #29, opened 2021-03-12
Parsing key name ending with "!" for supporting NOT searching

geniejhang:
This should be along with https://github.com/ControlSystemStudio/phoebus/pull/1829


shroffk (approved):
This is fine with me

shroffk:
I think we will have to fine tune this a bit, while there are use cases for searching for channels which do NOT have a particular tag or a particular property value... I am not sure we have a use case for NOT matching name.

In anycase we can sort this out before we update the documentation

PR #30, opened 2021-03-18
Create LICENSE

shroffk:


georgweiss:
Guess this is as good as any other open source license, right?

PR #32, opened 2021-04-27
Update README

mdavidsaver:
Updates based on some practice installs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ https://channelfinder.readthedocs.io/en/latest/
 
 ChannelFinder is a Java EE5 REST-style web service. The directory data is held in a ElasticSearch index.
 
+Collected installation recipes and notes may be found on [wiki pages](https://github.com/ChannelFinder/ChannelFinder-SpringBoot/wiki).
+
 * Prerequisites
 
   * JDK 8 or newer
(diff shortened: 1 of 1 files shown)


mdavidsaver:
@shroffk I know at one point I had some `curl` incantations to add/remove a CF tag through the service as a test of the elasticsearch install, but I can't find them now.  Do you remember how to do this?

mdavidsaver:
Also, rather than filling up the README with distribution/container/... specific instructions, I'll be adding wiki pages https://github.com/ChannelFinder/ChannelFinder-SpringBoot/wiki

mdavidsaver:
@shroffk What is going on with `src/main/resources/keystore/newcf.p12`?  It looks like this is being referenced by the default `application.properties`.  Is this being used somehow?

```
$ keytool -keystore src/main/resources/keystore/newcf.p12 -list
...
Your keystore contains 1 entry

cf, Jul 30, 2019, PrivateKeyEntry,
```

mdavidsaver:
I guess `newcf.p12` is an unused "dummy" key?

fyi. One way to create a truly empty keystore file (only 88 bytes) is:

```sh
keytool -keystore cf.keystore -storepass password -genkeypair -alias unused -dname "CN=unused"
keytool -keystore cf.keystore -storepass password -delete -alias unused
```

mdavidsaver:
API document needs to mention ".../resources/..." and give at least one example of a full URL.

```sh
$ curl http://localhost:8080/ChannelFinder/resources/tags
[]
```

shroffk (approved):

shroffk:
> API document needs to mention ".../resources/..." and give at least one example of a full URL.
> 
> ```shell
> $ curl http://localhost:8080/ChannelFinder/resources/tags
> []
> ```

I fixed this with a "example" at the start of the individual URL calls.

shroffk:
https://channelfinder.readthedocs.io/en/latest/api.html#web-service-urls-and-operations

I hope it makes sense...

shroffk:
Also, Thanks for adding the verification section... I should have documented it when I added it.
It is very useful to checking the state of the services and if it has a connection with elastic and the versions match etc...

PR #34, opened 2021-04-28
More documentation updates

mdavidsaver:
Fixup some sphinx warnings, add an example definition for the "Single Tag" payload, and add a demonstration to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,14 @@ $ curl http://localhost:8080/ChannelFinder
 }
 $ curl http://localhost:8080/ChannelFinder/resources/tags
 []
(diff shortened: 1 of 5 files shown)


shroffk (approved):

mdavidsaver:
I'll add definitions of all of the `Payload:` and `Returns:` which aren't `Scroll` (which I don't know).

mdavidsaver:
Ok, this is enough for me today.  I ended up doing some reorganization as well, and starting a page documenting `application.properties` keys (which I won't call "properties" to avoid confusion with, well, CF properties).  I also moved the introduction section to a separate page so that `index.rst` becomes a short list of links to other resources.

shroffk (approved):

PR #37, opened 2021-05-01
Add builtin Manager webpage

mdavidsaver:
To make it easier for new users to play around, and as a test of my understanding of the documentation, this PR adds a rudimentary management webpage to the CF service.  Allows creating/deleting tags, properties, and channels.  Assigning tags/properties to channels.  And simple querying.  This is not intended to be comprehensive, nor to be a serious substitute for other clients.  This will give someone with a newly installed server, and nothing else, an interface to explore the service.

The only java code change is to add some decorators to handle the CORS protocol mandated by modern browsers.


shroffk:
I know I should test this myself... but... thought it would be easier to ask.
The requests from these pages will ask for username and password and create a basic auth header?

mdavidsaver:
with firefox, I get prompted with the usual basic auth dialog.

PR #43, opened 2022-02-24
Auto init

shroffk:
Automatically create the indices if there are not already present


fredrik-soderberg (commented):
To create the indices automatically is a really nice feature. I have reviewed the code change and it looks ok to me.

PR #49, opened 2022-04-28
Add Dockerfile and docker-compose.

shroffk:
@fredrik-soderberg I tested this and it worked perfectly for me :)


shroffk:
I think the next steps would be to have the docker image checkout the latest master and build and run that

PR #55, opened 2022-07-19
Elastic8

shroffk:
Preparing for elastic 8


shroffk:
@tynanford @fredrik-soderberg I am ready to merge this

georgweiss (commented):
Sorry, cannot add any review at this point as I am ooo.

tynanford (approved):
Nice work @shroffk !

Took a bit to figure out how to get elastic 8 configured but after that CF started up and worked. My first test was with me manually populating ES with es8_mapping_definitions.sh . CF didn't seem to do that for me though when I tested with a blank ES instance. Isn't that supposed to be automatic now?

shroffk:
Yes it should be automatically creating the indices.
Let me double check

shroffk:
@tynanford 
I just pushed a commit which should generate the required indices

tynanford:
that fixed it thanks. Last thing, I see version 4.0.0 still in most places but also 4.0.5 in pom.xml

```
$ curl localhost:8080/ChannelFinder
{
    "name": "ChannelFinder Service",
    "version": "4.0.0",
    "elastic": {
....
```

https://github.com/ChannelFinder/ChannelFinderService/blob/elastic8/src/main/java/org/phoebus/channelfinder/InfoManager.java#L32

shroffk:
ya...
After this merge I want to sync up the next release version with the rest of the new epics services... so 4.7.0
I am hoping to fix all that as part of the next release

tynanford:
got it, thanks!

fredrik-soderberg (approved):
I have reviewed the code and also performed some basic testing with es8. The only issue that I have found is related to the docker files that needs to be updated. As we talked about earlier it might be best to fix the docker files after this merge.

Issue #57, opened 2022-08-25
Error when adding single tag or property to a PV via command line


tynanford:
I see this when trying to add a tag to a channel following this https://channelfinder.readthedocs.io/en/latest/api.html#add-tag-to-single-channel

```
$ curl http://localhost/ChannelFinder/resources/tags
[{"name":"testTag","owner":"admin","channels":[]}]

$ curl --basic -u admin:1234 -H 'Content-Type: application/json' -X PUT -d '{"name":"testTag", "owner":"admin"}' 
"http://localhost/ChannelFinder/resources/tags/testTag/testPV"
{"timestamp":"2022-08-25T17:22:22.114+0000","status":500,"error":"Internal Server Error",
"message":"Missing URI template variable 'channelName' for method parameter of type 
String","path":"/ChannelFinder/resources/tags/testTag/testPV"}
```

I think I have a fix, PR incoming

tynanford:
Fixed in https://github.com/ChannelFinder/ChannelFinderService/pull/56 which is just waiting on review before merge

PR #56, opened 2022-08-16
Updates and integrationtests adjust

thelarsjohansson:
Updates and integration tests
@shroffk

PR #58, opened 2022-08-25
Fix URI variable naming mismatch

tynanford:
Address https://github.com/ChannelFinder/ChannelFinderService/issues/57


shroffk:
thanks @tynanford I think this was fixed in the open PR by lars #56 
I am currently testing it before merging and releasing.

Fee free to test the open PR too.

tynanford:
oh nice I will close this PR then and if i get a chance to test #56 I'll comment there

PR #59, opened 2022-08-30
update all version number for CF and dependencies

shroffk:


thelarsjohansson:
@shroffk looks ok. I added another branch, patches_more, with suggestion for some documentation + more version number changes + some docker updates. (there were other occurrences of version number) It would be good to agree on release strategy before next version and release.

PR #65, opened 2022-09-29
Patches

shroffk:
Cleanup the imports
Fix the unnecessary initialization

Improve the port test cleanup


shroffk:
@kasemir The final commit adds support for case insensitive searching for property and tag name

kasemir (approved):

PR #68, opened 2022-12-19
AA integration

shroffk:


src/test/java/org/phoebus/channelfinder/AAChannelProcessorTest.java
```diff
@@ -0,0 +1,55 @@
...
+        ar.setPv("sim://testing");
+        ar.setSamplingParameters("monitor@1.0");
+
+        assertEquals(ar.getPv(), "sim://testing");
+        assertEquals(ar.getSamplingmethod(), "MONITOR");
+        assertEquals(ar.getSamplingperiod(), Float.valueOf("1.0"));
+
+        ar.setSamplingParameters("scan@.01");
+        assertEquals(ar.getSamplingmethod(), "SCAN");
+        assertEquals(ar.getSamplingperiod(), Float.valueOf(".01"));
```
mdavidsaver: Thinking of some other cases to test.  Some which I think should succeed:

```
MONITOR@1
MONITOR@0.01
MONITOR@1 ignore me
MONITOR@0.01 ignore me
```

And some which I think should fail (somehow):

```
INVALID
oops@
@1.0
@blah
INVALID@1
SCAN @1
SCAN@ 1.0
SCAN@1invalid
SCAN@invalid
MONTOR@1
```

tynanford:
Hi @shroffk 

This is really nice!!

A few comments
- It would be nice to have the archiver use PVAccess if desired and AFAIK that means the PV name requests to the AA need the pva:// prefix ( https://github.com/slacmshankar/epicsarchiverap/issues/105 ). I created a PR with a quick and dirty addition of a aa.pva property which would do that
- We use the named policies feature of AA so support for that would be good. I think that should be pretty easy to add, it would just be a "policy=<name_from_info_tag>" addition to the request JSON. We could also check the AA /mgmt/bpl/getPolicyList endpoint to confirm the policy name is in the AA before sending the request. What do you think?


Lastly, it seems in general like it is pretty easy to add PVs to the AA but more touchy to edit archiving parameters or delete the PV entirely. So in our current auto-archiving script at ALS we have found it beneficial to have that script be a more manual process. For example, during a maintenance period there might be a whole bunch of IOCs going up and down and some of those IOCs might be just for testing but happen to include reccaster. It's nice to run the auto-archiving script once after maintenance to avoid having a bunch of test and now non-existent PVs in the archiver. I don't have any ideas for how to address this, just throwing it out there. I guess one solution would be to take CF down at the start of maintenance and change aa.enable to false

PR #73, opened 2023-03-06
Log integrationtest maven

thelarsjohansson:
@shroffk 
Welcome to have a look! About log, integration test, adjusting pom.


shroffk:
great work @thelarsjohansson 

Just a request... in the future could we open smaller PR's which address individual issues. It makes it easier to review and revert if needed.

Elastic has started discouraging the use of the scroll API. I was thinking of deprecating it in ChannelFinder too.

thelarsjohansson:
Withdrawn, will be one pull request per kind of change.

PR #74, opened 2023-03-14
Adds support for using multiple archiver clusters.

jacomago:
Checks an extra property "archiver" to determine which archiver cluster to send the PV to.


src/main/java/org/phoebus/channelfinder/processors/AAChannelProcessor.java
```diff
@@ -11,10 +11,7 @@
 import org.springframework.web.reactive.function.client.WebClient;
 
 import java.net.URI;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
shroffk, line 14: We generally prefer to not use the wildcard imports

shroffk (approved):

PR #78, opened 2023-06-09
Collect log texts and use formatted strings in logging

thelarsjohansson:
Ensure consistent handling of loggers and log levels. 
Invoke log formatting conditionally. 
Collect log texts and use formatted strings in logging.

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/Application.java
+++ b/src/main/java/org/phoebus/channelfinder/Application.java
@@ -39,7 +39,7 @@ import org.springframework.util.FileCopyUtils;
 @SpringBootApplication
 public class Application implements ApplicationRunner {
 
-    static Logger logger = Logger.getLogger(Application.class.getName());
+    static final Logger logger = Logger.getLogger(Application.class.getName());
 
     public static void main(String[] args){
         // Set the java truststore used by channelfinder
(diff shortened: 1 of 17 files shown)


thelarsjohansson:
@shroffk 
@fredrik-soderberg 

Welcome to have a look!

shroffk:
Thank you for breaking up the PR

I think your changes and the TextUtil is great to standardize the logging and data representation.

Is there a particular reason we should not use `log.info(Supplier)` or `log.severe(Supplier)` and instead explicitly passing the level as an argument. Since the above would be a little less verbose and is the standard we are using in most other projects.

Also,
there are a few loggers which are configured with the `getCanonicalName()` (this is my mistake)... do you think we could fix that as part of this work of cleaning up the loggers.

thelarsjohansson:
The reason for logger.log(Level.INFO, "") instead of logger.info("") and similar, is that both were already used with the majority of uses being the former, i.e. logger.log(Level.INFO, ""). Then I standardized way of logging like that.

It would be easy to change to the latter, i.e. logger.info and similar.

shroffk:
> It would be easy to change to the latter, i.e. logger.info and similar.

+1 then, if it is all the same :)

thank you again for cleaning all this up.

shroffk:
Actually,
I think I found an argument for sticking with your setup.

logger.log(...) has better support for exception logging... so we should ignore my previous comment and keep things as they are

thelarsjohansson:
Ok. Then we leave it as is now. ( additional commit can be disregarded - same content as your code_cleanup branch ).

PR #82, opened 2023-07-19
Combined search and count

shroffk:


shroffk:
@jacomago 
Building on your latest PR I extracted the result class
Fixed some integration tests

Also,
split the the test the production properties

spring-shell.log
jacomago, line 1: What is this for?

src/main/java/org/phoebus/channelfinder/entity/SearchResult.java
jacomago, line 1: If we jump prerequiste jdk to 14+, this could be a record.

src/test/java/org/phoebus/channelfinder/ChannelRepositoryIT.java
```diff
@@ -180,39 +183,39 @@ public void testChannelsExist() {
...
-            e.printStackTrace();
-        } 
-    }
+//    /**
+//     * find all channels
+//     */
+//    @Test
+//    public void findAllXmlChannels() {
+//        Channel testChannel = new Channel("testChannel","testOwner",testProperties,testTags);
+//        Channel testChannel1 = new Channel("testChannel1","testOwner1",testProperties,testTags);
```
jacomago: Why is this now commented out?

jacomago (commented):
I think just the spring-shell.log to be deleted and the commented out test to be fixed or removed.

src/test/java/org/phoebus/channelfinder/ChannelRepositoryIT.java
shroffk: The final all had been initially added to satisfy the spring `CrudRepository ` interface.. however, for CF where there might be millions of channels this operation is not necessarily useful. Hence it is now not supported.

```
    @Override
    public Iterable<Channel> findAll() {
        throw new UnsupportedOperationException(TextUtil.FIND_ALL_CHANNELS_NOT_SUPPORTED);
    }
```

spring-shell.log
shroffk: arhg!!! this is accidental, removing it and updating the gitignore

src/main/java/org/phoebus/channelfinder/entity/SearchResult.java
shroffk: Let me ask around... I would be more inclined to move to JDK17 since it is a LTSed version.
However, my recent attempts to migrate AA to jdk17 have not been seamless

src/test/java/org/phoebus/channelfinder/ChannelRepositoryIT.java
shroffk: removed

shroffk:
Thanks @jacomago

jacomago (approved):
lgtm

PR #88, opened 2023-08-21
Adds testing to the github ci

jacomago:


jacomago:
Waiting for next release of testcontainers that includes fix for docker compose v2 https://github.com/testcontainers/testcontainers-java/pull/5608

src/main/java/org/phoebus/channelfinder/InfoManager.java
```diff
@@ -29,7 +29,7 @@
 @EnableAutoConfiguration
 public class InfoManager {
 
-    @Value("${channelfinder.version:4.7.0}")
+    @Value("${channelfinder.version:4.7.1}")
```
shroffk: This is just the default... the actual version number is picked up from the pom.xml. Maybe we should have a more obviously wrong default

# ChannelFinder version as defined in the pom file
channelfinder.version=@project.version@

src/site/sphinx/config.rst
```diff
@@ -6,7 +6,7 @@ Default configuration is `built in <https://github.com/ChannelFinder/ChannelFind
 which may be suitable for initial testing and evaluation.
 This may be overridden when starting the service.  eg. ::
 
-    java -jar ChannelFinder-4.7.0.jar -Dspring.config.location=file:./application.properties
+    java -jar ChannelFinder-4.7.1.jar -Dspring.config.location=file:./application.properties
```
shroffk: hmm...
maybe we replace this with a wildcard

 `   java -jar ChannelFinder-*.jar -Dspring.config.location=file:./application.properties`

shroffk (approved):

PR #89, opened 2023-08-21
Integrationtest documentation

thelarsjohansson:
Branch is about documentation on integrationtests with Docker. It includes instructions on how to run integration tests as well as tutorial for integration tests themselves. In addition, Maven surefire and failsafe plugins have been added that allow easier running of tests from command line. There is adjustment for port number used during test to not interfere with running instance of Elasticsearch.

Welcome to have a look!
@shroffk 
@jacomago

### DIFF
--- a/INTEGRATIONTEST_DOCKER_RUN.md
+++ b/INTEGRATIONTEST_DOCKER_RUN.md
@@ -1,0 +1,54 @@
+### Prerequisites
+
+##### Tools
+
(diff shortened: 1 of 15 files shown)


INTEGRATIONTEST_DOCKER_RUN.md
jacomago, line 1: I think this file and INTEGRATIONTEST_DOCKER_TUTORIAL.md should be in either src/test/resources/ or src/site/sphinx then they can be linked in the readme.md

PR #102, opened 2023-09-15
Add server compression by default for large responses

jacomago:


shroffk:
Looks good.
Is there any changes that would be needed on the clients?

jacomago:
It will only send if the client adds in the header that it supports compression. That is done by most browsers by default. Not sure if need an update to phoebus to enable using compression.

PR #104, opened 2023-09-18
Adds swagger-ui at /swagger-ui/ so clients can generate libraries.

jacomago:


shroffk:
Is it ok if I do the release without this PR?

Is there a reason we should use 1.X and not 2.X

jacomago:
Sure do a release without this. 

We'd need to also update to spring boot 3.x to use springdoc 2.x from what I understand.

PR #106, opened 2023-09-21
Makes sure to update the value of the property on the channels

### LINKED ISSUE #101: Property values sometimes overwritten with value from a different property
We have noticed that sometimes suddenly properties in CF become messed up and they show the value of a different property which is obviously wrong. For instance, suddenly the "archive" property of many PVs shows the hostName property value instead.

I think the issue is in this properties endpoint: https://github.com/ChannelFinder/ChannelFinderService/blob/master/src/main/java/org/phoebus/channelfinder/PropertyManager.java#L297

This line looks suspicious with the hard coded index of 0 but I haven't had time to debug it more: https://github.com/ChannelFinder/ChannelFinderService/blob/master/src/main/java/org/phoebus/channelfinder/PropertyManager.java#L350 

The backstory is we had a script populating the "archive" property for some PVs using this bulk update on /properties and found this script caused the problem. Switching the script to use the [bulk update on /channels](https://github.com/ChannelFinder/ChannelFinderService/blob/master/src/main/java/org/phoebus/channelfinder/ChannelManager.java#L321) stopped the issue from happening on "archive" but we still saw issues with the "pvStatus" property.

Looking at cfstore in recceiver, it uses the /properties update method for setting all channels Inactive on startup. After switching cfstore to use the /channels endpoint I have not seen the issue arise again for pvStatus. You can see the changes here: https://github.com/ChannelFinder/recsync/compare/master...tynanford:recsync:master

An unfortunate consequence of this problem is that it not only impacts the PVs in your query but many more. Probably because the method does a getChannels() on the property if it exists and includes those in the update.

jacomago:
to the value mentioned in the property payload uploaded.

Reproduces and fixes #101

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/PropertyManager.java
+++ b/src/main/java/org/phoebus/channelfinder/PropertyManager.java
@@ -5,6 +5,7 @@ import static org.phoebus.channelfinder.CFResourceDescriptors.PROPERTY_RESOURCE_
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


shroffk (approved):

shroffk:
Thank you,
This should address #101 

In the long run I wonder if there is a better way of doing this? is our reliance on nested documents making this more complicated that it should be

tynanford:
Thanks for the quick fix!

jacomago:
> Thank you, This should address #101
> 
> In the long run I wonder if there is a better way of doing this? is our reliance on nested documents making this more complicated that it should be

I thought the same thing fixing this. To me there is a natural hierachy that

Channels have properties and tags

So it seems weird to me to allow Properties to have channels or Tags to have channels. 

A more natural api to me would be:

A Property, Tag, and channel have a UID as well as the name_id which aren't returned from api, but are stored in db to make some of the queries simpler.

### /properties

POST /properties Add a new property 
PUT /properties/{propertyName} updates a property name only
GET /properties/{propertyName} returns a propertyName and Owner
DELETE /properties/{propertyName} Delete a property

### /tags

POST /tags Add a new property 
PUT /tags/{tagName} renames a tag only
GET /tags/{tagName} returns a tag and Owner
DELETE /tags/{tagName} deletes a tag

### /channels

POST /channels Add a new channel or channels, that can include a list of properties and tags that don't exist yet
PUT /channels Updates existing channels with new tags and property values and properties 
PUT /channels/{channelName} updates channelName and tags and properties of channel
GET /channels/{channelName} Gets a channel and its properties and tags

#### /channels sub objects

POST /channels/{channelName}/properties Replace properties list
PUT /channels/{channelName}/properties Add to properties list or update values of properties
POST /channels/{channelName}/properties/{propertyName} Add property to channel with value
PUT /channels/{channelName}/properties/{propertyName} Update value for a property
DELETE /channels/{channelName}/properties/{propertyName} Delete property from channel
GET /channels/{channelName}/properties/ Get properties of a channel including values
GET /channels/{channelName}/properties/{propertyName} Get single property of a channel

POST /channels/{channelName}/tags Replace tag list
PUT /channels/{channelName}/tags Add to tag list
POST /channels/{channelName}/tags/{tagName} Add tag to channel 
DELETE /channels/{channelName}/tags/{tagName} Delete tag from channel
GET /channels/{channelName}/tags/ Get tags of a channel 

#### /channels{q=Query}

POST /channels{q=Query}/properties Add a new property to queried channels 
PUT /channels{q=Query}/properties Update a already attached properties to queried channels to a new value
PUT /channels{q=Query}/properties/{propertyName} Update an already attached property to queried channels to a new value
DELETE /channels{q=Query}/properties Delete properties from queried channels 
DELETE /channels{q=Query}/properties/{propertyName} Delete property from queried channels 

POST /channels{q=Query}/tags Add a new tag or tags to queried channels 
DELETE /channels{q=Query}/tags Delete tags from queried channels 
DELETE /channels{q=Query}/tags/{tagName} Delete tag from queried channels 

Main thing is nesting is only done on /channels and POST, PUT /channels are the only bulk apis for updates. 

I think a new api could be written (Would be nice to use Spring Data). But channelfinder is pretty small so supporting the current nested documents is not that much work, so I'm not sure a new api is worth the effort. It would require either supporting two api versions at the same time, or tagging a chanelfinder version in channelfinder clients to check support.

tynanford:
> So it seems weird to me to allow Properties to have channels or Tags to have channels.

I agree, it makes the API/code more complicated without much clear benefit

PR #107, opened 2023-10-03
Add pause resume functionality to archiver functionality

jacomago:
Hi @tynanford @shroffk @mdavidsaver 

This is a proposal to add a new feature to the Channel Finder Archiver Appliance Processor.

The feature would automatically Pause, Resume or Archive PVs tagged with the "archive" infotag.

So in the end you would have:

     * name ArchiveStatus PVStatus - Archiver Appliance Action

     * PV Archived Active - Do Nothing
     * PV Paused Active - Resume
     * PV None Active - Archive
     * PV Archived Inactive - Pause
     * PV Paused Inactive - Nothing
     * PV None Inactive - Nothing

I need to do some more testing but I thought it would be good to get feedback on any problems there might be.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -182,6 +182,22 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>mockwebserver</artifactId>
+            <version>4.11.0</version>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>junit</groupId>
+                    <artifactId>junit</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId> <!-- Need for running tests on mac -->
+            <artifactId>netty-all</artifactId>
+        </dependency>
     </dependencies>
     <build>
         <!-- read properties from the pom file and add them to the application.properties -->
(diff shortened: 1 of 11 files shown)


jacomago:
One option is to add the possibility of pausing PVs that are being archived but are no longer being tagged.
Another is to update the policy in case the policy changes from what has been set.

tynanford:
Hi @jacomago , I think this makes sense but I'd like to test for a bit when you are ready to see how it works in practice. For instance, if an IOC reboots or if recceiver is rebooted then the PVs in the archiver will become paused for a little while and then resumed, I guess that is okay? Maybe a preference setting for aa_paused_on_inactive would be good to allow opting-out of this.

One concern I have is this adds a dependency on having recsync and the pvStatus fields be correct. We had some issues with this field in the past but it seems to have all been solved in your PR https://github.com/ChannelFinder/ChannelFinderService/pull/106 so that concern might not be valid anymore.

Overall, having some way to update the policies, pause PVs that no longer have tags, pause Inactive PVs, etc. would be really useful. We have already seen cases at ALS where PVs get added to the archiver mistakenly and have to be manually fixed. I like the idea of pausing PVs  that no longer have an archive tag in CF since otherwise those can be hard to find

jacomago:
@tynanford I updated the solution to have some configuration. So now there is a configuration parameter aa.auto_pause which can be set to:

none
pvStatus
archive
pvStatus,archive

Corresponding to whether to pause on change in pvStatus, or if the archive tag exists.

jacomago:
Thought of another idea, can probably add a timeout. i.e. how long a PV is inactive before pausing. I think can use the endpoint of the archiver to detemine last time there was a connection.

shroffk:
I am going to add this to the discussion topics for the next monthly developers meeting

PR #110, opened 2023-11-16
Update code coverage upload action

jacomago:


shroffk:
@jacomago do you have merge permissions? I think for changes like this you could merge the PR yourself

jacomago:
> @jacomago do you have merge permissions? I think for changes like this you could merge the PR yourself

Ah I don't have merge permissions.

PR #112, opened 2023-11-23
Code coverage it

thelarsjohansson:
Add ability to get code coverage for integration tests (Docker).
Update documentation for integration tests (Docker) - how to run, tutorial.
By default, code coverage for integration tests is disabled.
Additional cleanup for Sonar rules.


src/test/java/org/phoebus/channelfinder/docker/ChannelFinderChannelsIT.java
```diff
@@ -131,6 +135,28 @@ public static void tearDownObjects() {
         channel_c1_owner_o2 = null;
     }
 
+    @AfterAll
+    public static void extractJacocoReport() {
+        // extract jacoco report from container file system
+        //     stop jvm to make data available
```
jacomago, line 141: This method is repeated in all the docker IT Tests. It could be written once in ITUtil and called with the arguement of the class simple name.

jacomago (changes requested):

PR #113, opened 2023-11-23
Code coverage it

thelarsjohansson:
Add ability to get code coverage for integration tests (Docker).
Update documentation for integration tests (Docker) - how to run, tutorial.
By default, code coverage for integration tests is disabled.
Additional cleanup for Sonar rules.


src/test/java/org/phoebus/channelfinder/docker/ChannelFinderTagsIT.java
```diff
@@ -129,6 +133,28 @@ public static void tearDownObjects() {
...
 
+    @AfterAll
+    public static void extractJacocoReport() {
+        // extract jacoco report from container file system
+        //     stop jvm to make data available
+
+        if (!Boolean.FALSE.toString().equals(System.getProperty(ITUtil.JACOCO_SKIPITCOVERAGE))) {
+            return;
+        }
+
```
jacomago: This method is repeated in all the docker IT Tests. It could be written once in ITUtil and called with the arguement of the class simple name.

jacomago (changes requested):

src/test/java/org/phoebus/channelfinder/docker/ChannelFinderTagsIT.java
thelarsjohansson: Done. Also handled in similar way is ENVIRONMENT, i.e. instance of compose container.

jacomago (approved):

jacomago:
Looks good to me!

PR #114, opened 2023-11-24
Update to Java 17

jacomago:


shroffk:
It is nice to see how jdk17 would help get rid of a bunch of the boiler plate...not to mention the annoying xml binding dependencies.

shroffk:
trying to figure out how best to co-ordinate with all the collaborators on the move to JDK17

thelarsjohansson:
The "upgrade" path should be Spring Boot 2.7.18 + Java 17, before the step to Spring Boot 3 (e.g. 3.2.2) is made. In general, it's suitable and standard behavior to update so as to make the upcoming leap as small as possible. 

In addition

- Phoebus Olog already updated to Java 17
- 2.7.18 is update of minor version only

Therefore

- I recommend to update to Java 17 and Spring Boot 2.7.18 and use that as base for looking into Spring Boot 3.

shroffk (approved):

thelarsjohansson (approved):

thelarsjohansson:
ControlSystemStudio / phoebus and Olog / phoebus-olog have been updated to 17 (but not released thereafter).
It's logical step to do the same with ChannelFinder and allow for test and/or alpha/beta release if so desired.

If there are uncertainties, it would be useful to list those. It might be indication that specific test efforts are required.

shroffk:
Since we have the 4.7.3 alpha release in place I think we can merge this

PR #117, opened 2023-12-11
Update elastic search version

jacomago:


georgweiss (approved):
Any significant API changes spotted between 8.2.0 and 8.11.2?

jacomago:
> Any significant API changes spotted between 8.2.0 and 8.11.2?

Only needing to use the FieldValue in querys. https://github.com/ChannelFinder/ChannelFinderService/blob/c771291f4295f32bc1fd3af17fae268ebdc9c5b0/src/main/java/org/phoebus/channelfinder/ChannelRepository.java#L427

PR #122, opened 2024-01-22
Update version for spring framework boot to 2.7.18

thelarsjohansson:
Update spring framework boot version to 2.7.18.

Update to latest (and last) release on 2.7.* track in preparation of migration to Spring Boot 3.


jacomago (approved):
Looks good to me. Shame the epics test failed, but I don't think that is related to your change. @shroffk Should we remove it until we come up with a good use case?

PR #127, opened 2024-03-18
Improve configuration of connection to Elasticsearch

smarsching:
This PR introduces two major improvements:

1. It is now possible to use multiple URLs instead of a separately configured single hostname and port. This has two benefits: First, multiple hosts can be specified, enabling high-availability when using a multi-host Elasticsearch cluster. Second, the scheme can be specified, thus enabling the user to choose HTTPS instead of HTTP when connecting to the Elasticsearch cluster.
2. Credentials for connecting to the Elasticsearch cluster can be specified. This is required when security is enabled on the ES cluster. The supported methods are username / password, API key, and token-based authentication.

As a side effect, this PR also introduces three minor improvements:

1. The unused `clusterName` variable is removed from `ElasticConfig`.
2. The way how the `esInitialized` flag is used in `ElasticConfig` is changed, fixing a potential race condition.
3. Comments in `application.properties` that apparently were accidentally copied from the Elasticsearch server configuration file and were not relevant to the ChannelFinder service are removed.

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/ElasticConfig.java
+++ b/src/main/java/org/phoebus/channelfinder/ElasticConfig.java
@@ -138,7 +138,8 @@ public class ElasticConfig implements ServletContextListener {
         } else {
             client = currentClient;
         }
-        if (Boolean.parseBoolean(createIndices) && esInitialized.compareAndSet(false, true)) {
+        esInitialized.set(!Boolean.parseBoolean(createIndices));
+        if (esInitialized.compareAndSet(false, true)) {
             config.elasticIndexValidation(client);
         }
         return client;
--- a/src/main/java/org/phoebus/channelfinder/ElasticConfig.java
+++ b/src/main/java/org/phoebus/channelfinder/ElasticConfig.java
@@ -109,8 +109,7 @@ public class ElasticConfig implements ServletContextListener {
         } else {
             client = currentClient;
         }
-        esInitialized.set(!Boolean.parseBoolean(createIndices));
-        if (esInitialized.compareAndSet(false, true)) {
+        if (Boolean.parseBoolean(createIndices) && esInitialized.compareAndSet(false, true)) {
             config.elasticIndexValidation(client);
         }
         return client;
--- a/src/main/java/org/phoebus/channelfinder/ElasticConfig.java
+++ b/src/main/java/org/phoebus/channelfinder/ElasticConfig.java
@@ -17,8 +17,6 @@ package org.phoebus.channelfinder;
 import java.io.IOException;
 import java.io.InputStream;
 import java.text.MessageFormat;
-import java.util.Arrays;
-import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -75,7 +73,7 @@ public class ElasticConfig implements ServletContextListener {
     @Value("${elasticsearch.network.host:localhost}")
     private String host;
     @Value("${elasticsearch.host_urls:http://localhost:9200}")
-    private String hostUrls;
+    private HttpHost[] httpHosts;
     @Value("${elasticsearch.http.port:9200}")
     private int port;
     @Value("${elasticsearch.authorization.header:}")
@@ -114,14 +112,10 @@ public class ElasticConfig implements ServletContextListener {
             .addMixIn(Property.class, Property.OnlyProperty.class);
 
     private static ElasticsearchClient createClient(ElasticsearchClient currentClient, ObjectMapper objectMapper,
-                                                    List<String> hostUrls, String createIndices, ElasticConfig config) {
+                                                    HttpHost[] httpHosts, String createIndices, ElasticConfig config) {
         ElasticsearchClient client;
         if (currentClient == null) {
             // Create the low-level client
-            HttpHost[] httpHosts = new HttpHost[hostUrls.size()];
-            for (int i = 0; i < httpHosts.length; ++i) {
-                httpHosts[i] = HttpHost.create(hostUrls.get(i));
-            }
             RestClientBuilder clientBuilder = RestClient.builder(httpHosts);
             // Configure authentication
             if (!config.authorizationHeader.isEmpty()) {
@@ -155,15 +149,15 @@ public class ElasticConfig implements ServletContextListener {
 
     }
 
-    private List<String> getHostUrls() {
-        String localHostUrls = this.hostUrls;
+    private HttpHost[] getHttpHosts() {
+        HttpHost[] localHttpHosts = this.httpHosts;
         boolean hostIsDefault = host.equals("localhost");
-        boolean hostUrlsIsDefault = localHostUrls.equals("http://localhost:9200");
+        boolean hostUrlsIsDefault = localHttpHosts.length == 1 && localHttpHosts[0].equals(new HttpHost("http://localhost:9200"));
         boolean portIsDefault = (port == 9200);
         if (hostUrlsIsDefault) {
             if (!hostIsDefault || !portIsDefault) {
                 logger.warning("Specifying elasticsearch.network.host and elasticsearch.http.port is deprecated, please consider using elasticsearch.host_urls instead.");
-                localHostUrls = "http://" + host + ":" + port;
+                localHttpHosts = new HttpHost[] {new HttpHost("http://" + host + ":" + port)};
             }
         } else {
             if (!hostIsDefault) {
@@ -173,18 +167,18 @@ public class ElasticConfig implements ServletContextListener {
                 logger.warning("Only one of elasticsearch.host_urls and elasticsearch.http.port can be set, ignoring elasticsearch.http.port.");
             }
         }
-        return Arrays.asList(localHostUrls.split(","));
+        return localHttpHosts;
     }
 
     @Bean({ "searchClient" })
     public ElasticsearchClient getSearchClient() {
-        searchClient = createClient(searchClient, objectMapper, getHostUrls(), createIndices, this);
+        searchClient = createClient(searchClient, objectMapper, getHttpHosts(), createIndices, this);
         return searchClient;
     }
 
     @Bean({ "indexClient" })
     public ElasticsearchClient getIndexClient() {
-        indexClient = createClient(indexClient, objectMapper, getHostUrls(), createIndices, this);
+        indexClient = createClient(indexClient, objectMapper, getHttpHosts(), createIndices, this);
         return indexClient;
     }
 
--- a/src/main/java/org/phoebus/channelfinder/ElasticConfig.java
+++ b/src/main/java/org/phoebus/channelfinder/ElasticConfig.java
@@ -69,7 +69,6 @@ public class ElasticConfig implements ServletContextListener {
 
     private ElasticsearchClient searchClient;
     private ElasticsearchClient indexClient;
-    private static final AtomicBoolean esInitialized = new AtomicBoolean();
 
     @Value("${elasticsearch.network.host:localhost}")
     private String host;
@@ -138,8 +137,7 @@ public class ElasticConfig implements ServletContextListener {
         } else {
             client = currentClient;
         }
-        esInitialized.set(!Boolean.parseBoolean(createIndices));
-        if (esInitialized.compareAndSet(false, true)) {
+        if (Boolean.parseBoolean(createIndices)) {
             config.elasticIndexValidation(client);
         }
         return client;
--- a/src/main/java/org/phoebus/channelfinder/ElasticConfig.java
+++ b/src/main/java/org/phoebus/channelfinder/ElasticConfig.java
@@ -62,8 +62,6 @@ public class ElasticConfig implements ServletContextListener {
     private ElasticsearchClient indexClient;
     private static final AtomicBoolean esInitialized = new AtomicBoolean();
 
-    @Value("${elasticsearch.cluster.name:elasticsearch}")
-    private String clusterName;
     @Value("${elasticsearch.network.host:localhost}")
     private String host;
     @Value("${elasticsearch.http.port:9200}")
--- a/src/main/java/org/phoebus/channelfinder/ElasticConfig.java
+++ b/src/main/java/org/phoebus/channelfinder/ElasticConfig.java
@@ -17,6 +17,8 @@ package org.phoebus.channelfinder;
 import java.io.IOException;
 import java.io.InputStream;
 import java.text.MessageFormat;
+import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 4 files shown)


smarsching:
The issues that are reported by SonarCloud seem to be false positives. I cannot see anything that is wrong with the use of properties in the marked code and it works as expected.

jacomago:
@smarsching You're right. Sonarcloud doesn't seem to be good at evaluating those expressions. I'll try turning them off.

For the integration tests, you might need to update the docker-compose files to get them inline with your changes. And/or the application.properties file in the test resources folder.

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
```diff
@@ -62,12 +72,18 @@ public class ElasticConfig implements ServletContextListener {
     private ElasticsearchClient indexClient;
     private static final AtomicBoolean esInitialized = new AtomicBoolean();
 
-    @Value("${elasticsearch.cluster.name:elasticsearch}")
-    private String clusterName;
     @Value("${elasticsearch.network.host:localhost}")
     private String host;
+    @Value("${elasticsearch.host_urls:http://localhost:9200}")
+    private String hostUrls;
```
jacomago: SpeL expressions allow a list. So you should be able to extract an array here.

smarsching:
@jacomago I just took a look at the logs from the integration tests, and to me it does not look like the test failure is related to my changes. I specifically designed the changes, so that configuration through the old properties is still possible (so that users do not have to adapt their working configuration when upgrading).

To me, it rather looks like a problem with the Elasticsearch service in the test environment. For example:

    Search failed for {testProperty=[*]} Cause [es/search] failed: [search_phase_execution_exception] all shards failed”)

If you like, I can open a dummy PR without any real changes in order to confirm whether the integration tests go through for the unchanged code or not.

jacomago:
Another question I have is why do you want to do load balancing on Channelfinder, rather than using an inbetween load balancer? I can only think of doing this if you are using elastic for multiple applications, and then the setup of a load balancer would have the benefit of affecting every application. We currently have no performance problems with Channelfinder, so I'm not quite sure the use of multiple elastic nodes.

If you don't want to use a load balancer. I would rather you split up the adding authentication and adding multiple elastic nodes into multiple commits anyway.

In addition I think you should update the docker compose files and test properties to use the authentication, so we know it works in the tests.

smarsching:
> Another question I have is why do you want to do load balancing on Channelfinder, rather than using an inbetween load balancer? I can only think of doing this if you are using elastic for multiple applications, and then the setup of a load balancer would have the benefit of affecting every application. We currently have no performance problems with Channelfinder, so I'm not quite sure the use of multiple elastic nodes.

This is not about performance or load-balancing but about high availability. Yes, we run multiple applications on this ES cluster, but adding a load balancer in front of the cluster would make this load balancer the single point of failure, so there would be nothing gained from having a multi-node ES cluster.

Of course we could add a fail-over setup for the load-balancer, but this would render the whole setup even more complex. In general, we have a policy of implementing failover in the client, where reasonably possible, because it reduces complexity and makes it easier to find the underlying issue when a connection fails.

As the ES client library already supports this kind of configuration, just using it seemed like the easiest approach.

> If you don't want to use a load balancer. I would rather you split up the adding authentication and adding multiple elastic nodes into multiple commits anyway.

I already implemented this in multiple commits, it’s just a common branch because I needed all these changes combined for our setup. Authentication is implemented in https://github.com/ChannelFinder/ChannelFinderService/pull/127/commits/0eef1ab136afba1456c6b0da3cc506c73d8b3ba6 and the switch to URLs / supporting multiple ES hosts is implemented in https://github.com/ChannelFinder/ChannelFinderService/pull/127/commits/a75093866e2ca0dec949c1659d44b0dfb2e397b4. As written earlier, the second change preserves the ability to configure the host and port through the existing properties, so it should not affect users who don’t need it in any way.
 
> In addition I think you should update the docker compose files and test properties to use the authentication, so we know it works in the tests.

That’s a good idea. I will look into how authentication for ES can be enabled in the test environment.

smarsching:
I have added the changes, so that the test setup should use authentication in 42ff81f2fda20b62bb4f4b5ae0eff2e6c8ecd0fe.

.github/workflows/codecov.yml
```diff
@@ -33,6 +33,8 @@ jobs:
         elasticsearch-version: 8
         config: |
           http.port: 9200
+        env:
```
jacomago: Need to use the yml config, so I think something like:

config: |
  http.port: 9200
  elastic.password: super-secret

.github/workflows/maven.yml
```diff
@@ -70,6 +70,9 @@ jobs:
         elasticsearch-version: 8
         config: |
           http.port: 9200
+          xpack.security.enabled: true
+        env:
```
jacomago: Same here, enviroment variables aren't valid for the github action https://github.com/ankane/setup-elasticsearch/tree/v1

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
```diff
@@ -98,11 +114,32 @@ public int getES_QUERY_SIZE() {
...
 
     private static ElasticsearchClient createClient(ElasticsearchClient currentClient, ObjectMapper objectMapper,
-                                                    String host, int port, String createIndices, ElasticConfig config) {
+                                                    List<String> hostUrls, String createIndices, ElasticConfig config) {
         ElasticsearchClient client;
         if (currentClient == null) {
             // Create the low-level client
-            RestClient httpClient = RestClient.builder(new HttpHost(host, port)).build();
+            HttpHost[] httpHosts = new HttpHost[hostUrls.size()];
+            for (int i = 0; i < httpHosts.length; ++i) {
```
jacomago: You can use List.toArray here, or as mentioned us an array SpeL expression for elasticsearch.host_urls

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
```diff
@@ -98,11 +114,32 @@ public int getES_QUERY_SIZE() {
...
+                httpHosts[i] = HttpHost.create(hostUrls.get(i));
+            }
+            RestClientBuilder clientBuilder = RestClient.builder(httpHosts);
+            // Configure authentication
+            if (!config.authorizationHeader.isEmpty()) {
+                clientBuilder.setDefaultHeaders(new Header[] {new BasicHeader("Authorization", config.authorizationHeader)});
+                if (!config.username.isEmpty() || !config.password.isEmpty()) {
+                    logger.warning("elasticsearch.authorization_header is set, ignoring elasticsearch.usernamd and elasticsearch.password.");
```
jacomago: typo, usernamd should be username

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
```diff
@@ -98,11 +114,32 @@ public int getES_QUERY_SIZE() {
...
+                }
+            } else if (!config.username.isEmpty() || !config.password.isEmpty()) {
+                final CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
+                credentialsProvider.setCredentials(AuthScope.ANY, new UsernamePasswordCredentials(config.username, config.password));
+                clientBuilder.setHttpClientConfigCallback(new RestClientBuilder.HttpClientConfigCallback() {
+                    @Override
+                    public HttpAsyncClientBuilder customizeHttpClient(HttpAsyncClientBuilder httpClientBuilder) {
```
jacomago: Can this be a lambda?

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
```diff
@@ -111,22 +148,43 @@ private static ElasticsearchClient createClient(ElasticsearchClient currentClien
...
+        if (Boolean.parseBoolean(createIndices) && esInitialized.compareAndSet(false, true)) {
             config.elasticIndexValidation(client);
         }
         return client;
 
     }
+
+    private List<String> getHostUrls() {
+        String localHostUrls = this.hostUrls;
+        boolean hostIsDefault = host.equals("localhost");
```
jacomago, line 148: I think it's better to set the default of "host" and "port" to be empty strings. Then you are just checking if they are set rather than they are set to default. Then this check should be simpler.

src/test/resources/application_test.properties
```diff
@@ -91,6 +91,10 @@ elasticsearch.network.host: localhost
 #
 elasticsearch.http.port: 9200
```
jacomago: Can you unset elasticsearch.http.port and host, then set the host_urls. 
application_test.properties should use the most up to date settings.

jacomago (changes requested):
Can you also update the docker-compose.yml and the docker-compose-integrationtest.yml to include the authorization.

Thanks for the update.

.github/workflows/codecov.yml
smarsching: The respective configuration setting would be `bootstrap.password`, but this is a “secure” setting, so it cannot be set in the configuration file and rather has to be set in the keystore.

When using the Elasticsearch Docker image, the entrypoint script does exactly that when the container is started. Unfortunately, the `setup-elasticsearch` GitHub action does offer any way to set the bootstrap password and it does not offer any hook for running custom commands between extract ES and starting it either.

So, it looks like we cannot use authentication int the automated test setup and I am going to revert the last commit.

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
smarsching: That’s a very good point. I just changed it.

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
smarsching: I changed the code to directly create an array of `HttpHost` objects from the SpEL expression. This way, we do not have to do any conversion here any longer.

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
smarsching: Fixed

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
smarsching: Yes, it can be, so I changed it.

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
smarsching: I considered this option, but I discarded it for two reasons:

1. The port is converted to an integer, so Spring will thrown an exception when the associated property is the empty string.
2. When either of `elasticsearch.http.port` or `elasticsearch.network.host` is not the default value, I also need the value of the other property in order to build a valid URL, so I would still have to add the actual default values to the code, even if we decided to make the default value of the properties empty strings.

Thus, I think that checking whether the properties match the expected default values is better than using empty strings for the default values, even though I admit that I would have preferred the other solution, too.

src/test/resources/application_test.properties
smarsching: I made the change. I also removed all the commented out properties that do not actually exist in the configuration anyway (like I already did in `application.properties`).

While doing this, I noticed that a few properties in `application.properties` and `application_test.properties` were using the colon (`:`)  instead of the equal sign (`=`)  to separate property name and value, so I fixed this as well.

I think that (like the commented out properties), the colons got inadvertently introduced when copying lines from the Elasticsearch configuration file (which uses YAML syntax) to `application.properties` (which uses the Java properties file syntax).

smarsching:
@jacomago Thank you very much for your code review. I addressed the issues that you pointed out as far as I could with a reasonable amount of effort.

Regarding the authentication in the test environment, the only solution that I can see would be to write our own action for starting Elasticsearch instead of using https://github.com/ankane/setup-elasticsearch, but I do not think that just testing one small aspect (that authorization headers are added correctly) does not just the amount of work needed to completely replace that action with our own code, in particular from a maintenance perspective (updating the code for future versions of Elasticsearch).

jacomago:
@smarsching Figured out why the tests are failing. It is the new code. Seems like HttpHost when built from just a string, uses the string as the hostname rather than as a uri. 
Solution is to set the variable `elasticsearch.host_urls` to be a List<String> and use a conversion in `getHttpHosts` like so:

```java
HttpHost[] localHttpHosts = this.httpHosts.stream().map(HttpHost::create).toArray(HttpHost[]::new);
```

smarsching:
@jacomago Thanks for checking the code. You are right, I broke this when I changed the code to directly inject the property into an array of HttpHost objects.

I didn’t notice it because I didn’t update our ChannelFinder instance to use the newest version of the code after this change. Now, I have updated our ChannelFinder to use the latest version of this branch and it seems to be finde, so I am reasonably sure that this problem is resolved.

jacomago:
@smarsching I think maybe the problem might have been with the version of elasticsearch. I've updated it on master, so if you rebase or merge in the changes then hopefully the CI tests will pass. 
I think the CI tests should be passable since other actions have passed recently.

smarsching:
@jacomago I just rebased my changes on top of the current `master` branch, but the checks still seem to fail.

Interestingly, when I run the checks locally with [act](https://nektosact.com/), they also fail for both the `master` and `es-connection-improvements` branch.

Looking at the log output, the cause of the issues seems to have changed though: It seems like most or all tests fail now, but they consistently fail with `NoClassDefFound com/fasterxml/jackson/core/StreamReadConstraints`. I will try to find the cause of this exception.

smarsching:
@jacomago I managed to fix the problem with Jackson: The cause was that different versions of jackson-core (2.14.2) and jackson-databind (2.16.0) were used. This problem got introduced when I rebased my changes on the latest version of the `master` branch.

After fixing this issue in my branch, we are now back to the original situation, where only a few tests fail. I created a separate PR #130, that only fixes the Jackson issue. If the CI checks succeed for that PR, we know that some of the other changes I made must be the cause. If the CI fails for that PR as well, the problems must also be present in the `master` branch.

jacomago:
> After fixing this issue in my branch, we are now back to the original situation, where only a few tests fail.

Just to note that the tests that fail are the ones that rely on the elasticsearch setup by the action. The tests that passed create a new elastic docker container for each test (a bit complicated I know, plans are to merge the tests).

smarsching:
@jacomago Okay, so the cause for the problems were changes that I made as part of this PR, in particular the change that I made to the initialization logic.

I changed:

```java
        esInitialized.set(!Boolean.parseBoolean(createIndices));
        if (esInitialized.compareAndSet(false, true)) {
            config.elasticIndexValidation(client);
        }
```

to

```java
         if (Boolean.parseBoolean(createIndices) && esInitialized.compareAndSet(false, true)) {
             config.elasticIndexValidation(client);
         }
```

The difference is that my version only runs `config.elasticIndexValidation(client)` once, while the original version runs it every time `createClient(…)` is callled. It seems that for some reason, calling this method mulitple times is needed, at least in the test context.

Under this circumstances, having the `esInitialized` flag does not make any sense, so now I made changes to remove it completely:

```java
        if (Boolean.parseBoolean(createIndices)) {
            config.elasticIndexValidation(client);
        }
```

This way, it at least is clear that this method might be called more than once.

With these checks now succeeding, are there any other issues that need to be addressed before this PR can be merged?

jacomago:
Awesome! I will merge it in.

jacomago (approved):

smarsching:
@jacomago Thanks for your help in figuring this issue out.

PR #135, opened 2024-04-11
Disable listening on ipv6 in the docker compose files.

jacomago:
Should reduce the erroneous logs from the pvAccess server in docker tests.

See https://github.com/ControlSystemStudio/phoebus/blob/5a69830bd5b4be7a23a1359dcb49a62de61049cb/core/pva/src/main/java/org/epics/pva/PVASettings.java#L130 for details of EPICS_PVAS_INTF_ADDR_LIST

### DIFF
--- a/docker-compose-integrationtest.yml
+++ b/docker-compose-integrationtest.yml
@@ -40,7 +40,7 @@ services:
       embedded_ldap.enabled: "false"
       demo_auth.enabled: "true"
       skipITCoverage: "true"
-      EPICS_PVA_ENABLE_IP6: "false"
+      EPICS_PVAS_INTF_ADDR_LIST: "0.0.0.0"
     command: >
       /bin/bash -c "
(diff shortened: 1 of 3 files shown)


jacomago:
@thelarsjohansson This one might fix the test failure in https://github.com/ChannelFinder/ChannelFinderService/pull/122

tynanford (approved):
unrelated to this PR but is there a reason to expose elasticsearch port 9200 in the main docker-compose.yml? I think it's better to keep that only in the internal docker network and only expose the CF API

jacomago:
> unrelated to this PR but is there a reason to expose elasticsearch port 9200 in the main docker-compose.yml? I think it's better to keep that only in the internal docker network and only expose the CF API

I think it was only kept for debugging purposes. I'll remove it in another pull request.

thelarsjohansson (approved):

Issue #139, opened 2024-04-23
Can't load any channel above 10000 - Error 500


kingspride:
As mentioned in the recsync issues,
(https://github.com/ChannelFinder/recsync/issues/86)

I have a wierd behavior of my channelfinder 4.7.1.
according to `/resources/channels/count`, I have 24,753 in total.

my application.properties has a size limit of 10,000.

so, I tried to get the remaining channels after the 10,000th: `/resources/channels?~size=1&~from=10000`
which resulted in an Error 500.
the log says:
```
2024-04-23 13:38:28.910 ERROR 1206469 --- [nio-8443-exec-8] o.p.channelfinder.ChannelRepository      : Search failed for: {~size=[1], ~from=[10000]}

co.elastic.clients.elasticsearch._types.ElasticsearchException: [es/search] failed: [search_phase_execution_exception] all shards failed
	at co.elastic.clients.transport.rest_client.RestClientTransport.getHighLevelResponse(RestClientTransport.java:281) ~[elasticsearch-java-8.2.0.jar!/:na]
	at co.elastic.clients.transport.rest_client.RestClientTransport.performRequest(RestClientTransport.java:147) ~[elasticsearch-java-8.2.0.jar!/:na]
	at co.elastic.clients.elasticsearch.ElasticsearchClient.search(ElasticsearchClient.java:1502) ~[elasticsearch-java-8.2.0.jar!/:na]
	at org.phoebus.channelfinder.ChannelRepository.search(ChannelRepository.java:488) ~[classes!/:na]
	at org.phoebus.channelfinder.ChannelRepository$$FastClassBySpringCGLIB$$ac7847c2.invoke(<generated>) ~[classes!/:na]
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218) ~[spring-core-5.3.22.jar!/:5.3.22]
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:793) ~[spring-aop-5.3.22.jar!/:5.3.22]
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163) ~[spring-aop-5.3.22.jar!/:5.3.22]
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:763) ~[spring-aop-5.3.22.jar!/:5.3.22]
[...]

PR #141, opened 2024-04-24
Issue #139

shroffk:
#139 
Set the max search window using ChannelFinder preferences
Throw a bad request exception when the search size+from exceeds the allowed window

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/ChannelRepository.java
+++ b/src/main/java/org/phoebus/channelfinder/ChannelRepository.java
@@ -415,7 +415,7 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
         Integer finalSize = builtQuery.size;
         Integer finalFrom = builtQuery.from;
 
-        if(builtQuery.size + builtQuery.from >= esService.getES_MAX_RESULT_WINDOW_SIZE()) {
+        if(builtQuery.size + builtQuery.from > esService.getES_MAX_RESULT_WINDOW_SIZE()) {
             String message = MessageFormat.format(TextUtil.SEARCH_FAILED_CAUSE,
                     searchParameters,
                     "Max search window exceeded, use the " + CFResourceDescriptors.SCROLL_RESOURCE_URI + " api.");
--- a/src/main/java/org/phoebus/channelfinder/ChannelRepository.java
+++ b/src/main/java/org/phoebus/channelfinder/ChannelRepository.java
@@ -412,11 +412,17 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
      */
     public SearchResult search(MultiValueMap<String, String> searchParameters) {
         BuiltQuery builtQuery = getBuiltQuery(searchParameters);
+        Integer finalSize = builtQuery.size;
+        Integer finalFrom = builtQuery.from;
+
+        if(builtQuery.size + builtQuery.from >= esService.getES_MAX_RESULT_WINDOW_SIZE()) {
+            String message = MessageFormat.format(TextUtil.SEARCH_FAILED_CAUSE,
+                    searchParameters,
+                    "Max search window exceeded, use the " + CFResourceDescriptors.SCROLL_RESOURCE_URI + " api.");
+            throw new ResponseStatusException(HttpStatus.BAD_REQUEST, message);
+        }
 
         try {
-            Integer finalSize = builtQuery.size;
-            Integer finalFrom = builtQuery.from;
-
             SearchRequest.Builder searchBuilder = new SearchRequest.Builder();
             searchBuilder.index(esService.getES_CHANNEL_INDEX())
                             .query(builtQuery.boolQuery.build()._toQuery())
--- a/src/main/java/org/phoebus/channelfinder/ChannelRepository.java
+++ b/src/main/java/org/phoebus/channelfinder/ChannelRepository.java
@@ -275,7 +275,7 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
             SearchRequest.Builder searchBuilder = new SearchRequest.Builder()
                     .index(esService.getES_CHANNEL_INDEX())
                     .query(IdsQuery.of(q -> q.values(ids))._toQuery())
-                    .size(10000)
+                    .size(esService.getES_QUERY_SIZE())
                     .sort(SortOptions.of(s -> s.field(FieldSort.of(f -> f.field("name")))));
             SearchResponse<Channel> response = client.search(searchBuilder.build(), Channel.class);
             return new HashSet<>(response.hits()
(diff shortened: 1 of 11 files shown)


src/main/java/org/phoebus/channelfinder/ElasticConfig.java
```diff
@@ -91,8 +95,10 @@ public class ElasticConfig implements ServletContextListener {
     private String ES_PROPERTY_INDEX;
     @Value("${elasticsearch.channel.index:channelfinder}")
     private String ES_CHANNEL_INDEX;
-    @Value("${elasticsearch.query.size}")
+    @Value("${elasticsearch.query.size:10000}")
     private int ES_QUERY_SIZE;
+    @Value("${elasticsearch.maxResultWindow.size:10000}")
```
jacomago: Do we need both of these? Can we not combine them?

src/main/resources/application.properties
```diff
@@ -102,6 +102,9 @@ elasticsearch.channel.index = channelfinder
 # maximum query result size
 elasticsearch.query.size = 10000
 
+# WARNING this changes the elastic settings. UPDATE  with care.
+elasticsearch.maxResultWindow.size = 10000
```
jacomago: You should update the test application.properties.

src/main/java/org/phoebus/channelfinder/ChannelRepository.java
```diff
@@ -412,11 +412,17 @@ public void deleteAll() {
      */
     public SearchResult search(MultiValueMap<String, String> searchParameters) {
         BuiltQuery builtQuery = getBuiltQuery(searchParameters);
+        Integer finalSize = builtQuery.size;
+        Integer finalFrom = builtQuery.from;
+
+        if(builtQuery.size + builtQuery.from >= esService.getES_MAX_RESULT_WINDOW_SIZE()) {
+            String message = MessageFormat.format(TextUtil.SEARCH_FAILED_CAUSE,
+                    searchParameters,
+                    "Max search window exceeded, use the " + CFResourceDescriptors.SCROLL_RESOURCE_URI + " api.");
```
jacomago, line 421: Could you add a test case to ChannelRepositoryIT that checks that this is returned?

jacomago (changes requested):

jacomago:
A couple of other things missing that I thought of. Can you update the ChannelManager.query javadoc (this is what appears in the swagger). And update the site/api.rst (which I think we should move into the javadoc as a different PR).

src/main/resources/application.properties
shroffk: sure

src/main/java/org/phoebus/channelfinder/ChannelRepository.java
shroffk: I am working in the tests for this...
I am having the same issues that the github CI build is having with respect to running the IT tests.
I think I have an issue ticket... I am trying to run the tests from intellij using 
`failsafe:integration-test -DskipITs=false`

jacomago: I usually use the full method from the github actions.

`mvn clean install test-compile failsafe:integration-test failsafe:verify --batch-mode --fail-at-end -Djacoco.skip=true -DskipITs=false -DskipITCoverage=true`

 I keep meaning to make a PR to simplify it....

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
```diff
@@ -204,13 +213,16 @@ private void validateIndex(ElasticsearchClient client, String esIndex, String ma
         try (InputStream is = ElasticConfig.class.getResourceAsStream(mapping)) {
             BooleanResponse exits = client.indices().exists(ExistsRequest.of(e -> e.index(esIndex)));
             if(!exits.value()) {
-
                 CreateIndexResponse result = client.indices().create(
                         CreateIndexRequest.of(
-                                c -> c.index(esIndex).withJson(is)));
-                logger.log(Level.INFO, () -> MessageFormat.format(TextUtil.CREATED_INDEX_ACKNOWLEDGED, esIndex, result.acknowledged()));
+                                c -> c.index(ES_CHANNEL_INDEX)
```
jacomago: Should be esIndex

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
```diff
@@ -204,13 +213,16 @@ private void validateIndex(ElasticsearchClient client, String esIndex, String ma
...
+                                        .withJson(is)
+                                        .settings(IndexSettings.of(builder -> builder.maxResultWindow(ES_MAX_RESULT_WINDOW_SIZE)))));
+                logger.log(Level.INFO, () -> MessageFormat.format(TextUtil.CREATED_INDEX_ACKNOWLEDGED, ES_CHANNEL_INDEX, result.acknowledged()));
             }
-
+            PutIndicesSettingsResponse response = client.indices()
+                    .putSettings(PutIndicesSettingsRequest.of(builder -> builder.index(ES_CHANNEL_INDEX).settings(IndexSettings.of(i -> i.maxResultWindow(ES_MAX_RESULT_WINDOW_SIZE)))));
+            logger.log(Level.INFO, () -> MessageFormat.format(TextUtil.UPDATE_INDEX_ACKNOWLEDGED, ES_CHANNEL_INDEX, response.acknowledged()));
```
jacomago: Should be esIndex not ES_CHANNEL_INDEX

jacomago (changes requested):
Might be the failure reason.

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
shroffk: Sorry,
An artifact of the merge conflict resolution...

src/main/java/org/phoebus/channelfinder/ChannelRepository.java
jacomago: I think this should be < not <=

shroffk: Sorry, 
Where should it be `<`
Currently, there is just one comparison I see size+from >= window size

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
shroffk: hmm... 
I guess we could, since this setting updates the actual index settings I thought it would be safer to have it explicitly defined... I am happy to be convinced the other way if you think I am being over cautious.

src/main/java/org/phoebus/channelfinder/ChannelRepository.java
jacomago: Sorry. Yes. That comparison should be >

shroffk: Great!! and you are correct, `>` should be used to determine when the elastic query will fail

src/main/java/org/phoebus/channelfinder/ElasticConfig.java
jacomago: I think it's better to have them combined for simplicity. Then can split if someone has a usecase otherwise. I don't see much reason to have them different amounts as you can always override with the ~size when you want smaller than the window.

jacomago (approved):

Issue #144, opened 2024-06-15
Allow for multiple default archivers in AA plugin


tynanford:
It would be useful for our site if the `aa.default_alias` property is a list instead of single archiver alias. We have two appliance instances and would like PVs by default to be placed in each instance.

It wouldn't be much more effort to allow comma separated lists in the `archiver` info tag as well but that isn't needed for us. Does that make sense to add or not?

shroffk:
I see no harm in the default_alias being a list. ( it does complicate things... since now you have to match a list of aliases to the list of aa urls... should we just support something like "all" defined url's )

regarding the archiver info tag... I think it should support the same features that the default alias supports. i.e. a single archiver url or a list ( I am guessing we only need to make changes to the processor )

tynanford:
> should we just support something like "all" defined url's

That could be a nice feature but it doesn't fit our use case at ALS. I am thinking to keep that a separate issue/PR than this one unless you think I should add it

with #145 the archiver tag can now be a list of archivers too

PR #145, opened 2024-06-15
Change aa default_alias to list of strings

tynanford:
https://github.com/ChannelFinder/ChannelFinderService/issues/144

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/processors/AAChannelProcessor.java
+++ b/src/main/java/org/phoebus/channelfinder/processors/AAChannelProcessor.java
@@ -55,7 +55,7 @@ public class AAChannelProcessor implements ChannelProcessor {
     @Value("#{${aa.urls:{'default': 'http://localhost:17665'}}}")
     private Map<String, String> aaURLs;
     @Value("${aa.default_alias:default}")
-    private List<String> defaultArchiver;
+    private List<String> defaultArchivers;
     @Value("${aa.pva:false}")
     private boolean aaPVA;
     @Value("${aa.archive_property_name:archive}")
@@ -113,7 +113,7 @@ public class AAChannelProcessor implements ChannelProcessor {
                         .map(Property::getValue)
                         .collect(Collectors.toList()))
                         .filter(list -> !list.isEmpty())
-                        .orElse(defaultArchiver);
+                        .orElse(defaultArchivers);
                 for(String archiverAlias : archivers) {
                     try {
                         addChannelChange(channel, aaArchivePVS, policyLists, archiveProperty, archiverAlias);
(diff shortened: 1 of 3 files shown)


src/main/java/org/phoebus/channelfinder/processors/AAChannelProcessor.java
```diff
@@ -55,7 +55,7 @@ public class AAChannelProcessor implements ChannelProcessor {
     @Value("#{${aa.urls:{'default': 'http://localhost:17665'}}}")
     private Map<String, String> aaURLs;
     @Value("${aa.default_alias:default}")
-    private String defaultArchiver;
+    private List<String> defaultArchiver;
```
jacomago: Good to rename to "defaultArchivers"

src/main/java/org/phoebus/channelfinder/processors/AAChannelProcessor.java
```diff
@@ -108,10 +108,18 @@ public long process(List<Channel> channels) throws JsonProcessingException {
...
-                    logger.log(Level.WARNING, String.format("Failed to process %s", channel), e);
+                List<String> archivers = Optional.of(channel.getProperties().stream()
+                        .filter(xmlProperty -> archiverPropertyName.equalsIgnoreCase(xmlProperty.getName()))
+                        .map(Property::getValue)
+                        .collect(Collectors.toList()))
+                        .filter(list -> !list.isEmpty())
+                        .orElse(defaultArchiver);
+                for(String archiverAlias : archivers) {
+                    try {
+                        addChannelChange(channel, aaArchivePVS, policyLists, archiveProperty, archiverAlias);
```
jacomago: This can be done inside the stream with a .forEach, so then there should be no need to do a collect.

jacomago (changes requested):

src/main/java/org/phoebus/channelfinder/processors/AAChannelProcessor.java
tynanford: updated

src/main/java/org/phoebus/channelfinder/processors/AAChannelProcessor.java
tynanford: okay I updated this and now you can also specify a comma separated list for the `archiver` info tag

shroffk:
@tynanford Is this ready? or close to ready?

shroffk (approved):

tynanford:
Hi @shroffk yes this is ready. @jacomago does it look okay to you?

jacomago (approved):
Looks good to me!

PR #149, opened 2024-06-25
Bulk archiver status for Channel update greater than 8k

### LINKED ISSUE #148: issue in AA processor for large list of pvs
The archiver status end point can only handle so many pvs at once. If large IOCs restart, or a recceiver restarts can suddenly get a lot of PVs being processed. 

Should batch in groups of 100 or something. @shroffk Maybe this should be done at the general processor level?

jacomago:
Refactors the AAChannelProcessor into multiple classes as it was getting quite big.
Split up requests for archiving status when there are a large number of PVs.
Doesn't split up when using the newer POST Api on Archiver version after 1.1.0

Fixes #148

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -122,7 +122,6 @@ aa.enabled=true
 aa.pva=false
 aa.archive_property_name=archive
 aa.archiver_property_name=archiver
-aa.version:1.1.0
 
 # Set the auto pause behaviour
 #
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -112,7 +112,7 @@ channelfinder.version=@project.version@
 
 ############################## REST Logging ###############################
 # DEBUG level will log all requests and responses to and from the REST end points
-logging.level.org.springframework.web.filter.CommonsRequestLoggingFilter=INFO
+logging.level.org.springframework.web.filter.CommonsRequestLoggingFilter=DEBUG
 
 ################ Archiver Appliance Configuration Processor #################
 aa.urls={'default': 'http://localhost:17665'}
@@ -122,6 +122,7 @@ aa.enabled=true
 aa.pva=false
 aa.archive_property_name=archive
 aa.archiver_property_name=archiver
+aa.version:1.1.0
 
 # Set the auto pause behaviour
 #
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -112,7 +112,7 @@ channelfinder.version=@project.version@
 
 ############################## REST Logging ###############################
 # DEBUG level will log all requests and responses to and from the REST end points
-logging.level.org.springframework.web.filter.CommonsRequestLoggingFilter=DEBUG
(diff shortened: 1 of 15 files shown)


src/main/resources/application.properties
```diff
@@ -112,15 +112,16 @@ channelfinder.version=@project.version@
 
 ############################## REST Logging ###############################
 # DEBUG level will log all requests and responses to and from the REST end points
-logging.level.org.springframework.web.filter.CommonsRequestLoggingFilter=INFO
+logging.level.org.springframework.web.filter.CommonsRequestLoggingFilter=DEBUG
 
 ################ Archiver Appliance Configuration Processor #################
-aa.urls={'default': 'http://localhost:17665'}
+aa.urls={'default': 'http://skybreweru.cslab.esss.lu.se:17665'}
```
tynanford: Let's keep this default as localhost

tynanford:
Overall looks really nice. Is there a way to get the version from the archiver API instead of having to place it in the properties file?

Right now when I populate our appliance from scratch with recsync/CF, it takes something like 8-12 hours for the appliance to get all ~50K PVs being archived. They all make it into the archive request MySQL table quickly but then slowly trickle their way into the archiver. Is this the same issue you are seeing and addressing with the PR?

jacomago:
> Overall looks really nice. Is there a way to get the version from the archiver API instead of having to place it in the properties file?
>
Yes, I'll try add that.
 
> Right now when I populate our appliance from scratch with recsync/CF, it takes something like 8-12 hours for the appliance to get all ~50K PVs being archived. They all make it into the archive request MySQL table quickly but then slowly trickle their way into the archiver. Is this the same issue you are seeing and addressing with the PR?

No this problem is when you have an IOC with 200+ PVs wanting to be archived. Then when channelfinder fetches the status of the PVs, it does so using a query request which is a really big url. Default Tomcat setttings on Archiver Appliance limit is 8k characters for the size of the url before rejecting. 

The slowness of connecting to PVs is mostly due to the archiver setting (archappl.properties) org.epics.archiverappliance.mgmt.MgmtRuntimeState.archivePVWorkflowBatchSize which throttles how many PVs an appliance can start archiving at once. We put this up to 10000 when we migrated hardware.

tynanford:
Ah I see, thanks!

src/main/resources/application.properties
```diff
@@ -112,7 +112,7 @@ channelfinder.version=@project.version@
 
 ############################## REST Logging ###############################
 # DEBUG level will log all requests and responses to and from the REST end points
-logging.level.org.springframework.web.filter.CommonsRequestLoggingFilter=INFO
```
tynanford: sorry to be picky but I would leave the default here to INFO

tynanford (approved):
This looks good to me. The only thought would be how it impacts people running the archiver on a non-tagged archiver commit. Then the version check probably won't find a match in the list.

I don't think it's a reason not to merge the PR though, since the behavior then would be the same as before

PR #153, opened 2024-08-06
Enable reproducible builds

minijackson:
Hello, and thanks for this project!

This PR enables reproducible builds, which are useful for checking that files were not tampered with.

By setting the timestamp of produced files, we enable artifacts to be bit-for-bit reproducible from one build to another.

Tested by running:

    mvn artifact:check-buildplan
    mvn clean install
    mvn clean artifact:compare

Done with the help of these guides:

- https://reproducible-builds.org/docs/jvm/
- https://maven.apache.org/guides/mini/guide-reproducible-builds.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -35,6 +35,7 @@
     </developers>
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.build.outputTimestamp>2024-01-01T00:00:00Z</project.build.outputTimestamp>
         <spring.boot-version>2.7.3</spring.boot-version>
         <elasticsearch.version>8.11.2</elasticsearch.version>
         <junit-jupiter.version>5.10.0</junit-jupiter.version>


pom.xml
```diff
@@ -35,6 +35,7 @@
     </developers>
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.build.outputTimestamp>2024-01-01T00:00:00Z</project.build.outputTimestamp>
```
jacomago, line 38: Not sure we want an explicit time stamp. Would be nice if it was the git commit from https://maven.apache.org/guides/mini/guide-reproducible-builds.html or just the timestamp from the build time.

jacomago (changes requested):

pom.xml
minijackson: The timestamp from the build time makes the build not reproducible, since two builds of the same source code will give two different timestamps.

For the timestamp from the git commit, I don't mind changing the PR, but it will not work if the user downloads ZIP files from GitHub, which a certain number of build system does by default.

jacomago: I think getting the timestamp from the git commit, and then if it doesn't exist use a default timestamp makes sense. Otherwise the timestamp will most of the time be nonsensical.

minijackson: Sounds good to me! However I'm not sure how to implement it, I'm not a Java person. If you know how, please tell me and I'll update the PR. Else I might have to ask around.

shroffk: We could look into this plugin 
https://github.com/git-commit-id/git-commit-id-maven-plugin

we don't need to access this info in code or generate a properties file... but simply try to get it as a maven property

shroffk:
```
    <build>
        <plugins>
            <plugin>
                <groupId>io.github.git-commit-id</groupId>
                <artifactId>git-commit-id-maven-plugin</artifactId>
                <version>9.0.1</version>
                <executions>
                    <execution>
                        <id>get-the-git-infos</id>
                        <goals>
                            <goal>revision</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <dotGitDirectory>${project.basedir}/.git</dotGitDirectory>
                    <prefix>git</prefix>
                    <verbose>false</verbose>
                    <generateGitPropertiesFile>false</generateGitPropertiesFile>
                </configuration>
            </plugin>
        </plugins>
        <finalName>${artifactId}-${git.commit.id.abbrev}</finalName>
    </build>
```

you can use any one of these properties
```
    "git.branch": "master",
    "git.build.host": "LPS-164306",
    "git.build.time": "2024-09-09T13:21:03-04:00",
    "git.build.user.email": "kunalshroff9@gmail.com",
    "git.build.user.name": "Kunal Shroff",
    "git.build.version": "1.0-SNAPSHOT",
    "git.closest.tag.commit.count": "",
    "git.closest.tag.name": "",
    "git.commit.author.time": "2024-09-09T09:16:28-04:00",
    "git.commit.committer.time": "2024-09-09T09:16:28-04:00",
    "git.commit.id": "092b2e3d56621e010ddbf20d5899a68068d4f87f",
    "git.commit.id.abbrev": "092b2e3",
    "git.commit.id.describe": "092b2e3-dirty",
    "git.commit.id.describe-short": "092b2e3-dirty",
    "git.commit.message.full": "testing",
    "git.commit.message.short": "testing",
    "git.commit.time": "2024-09-09T09:16:28-04:00",
    "git.commit.user.email": "shroffk@bnl.gov",
    "git.commit.user.name": "Shroff"
```

shroffk:
https://github.com/ChannelFinder/ChannelFinderService/pull/159

PR #154, opened 2024-09-02
Test multiple PVs being processed and fix bug in processsing multiple pvs

jacomago:


shroffk (approved):

shroffk:
+1

we might want to plan for checks to ensure that the URL length when using a list of PVs does not exceed 2048 chars

PR #159, opened 2024-10-01
Reproducible builds

shroffk:
Based on similar work as:

https://github.com/ControlSystemStudio/phoebus/pull/3153

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -364,6 +364,9 @@
                     <generateGitPropertiesFile>true</generateGitPropertiesFile>
                     <generateGitPropertiesFilename>${project.build.outputDirectory}/git.properties</generateGitPropertiesFilename>
                     <format>json</format>
+                    <excludeProperties>
+                        <excludeProperty>^git.build.*$</excludeProperty>
+                    </excludeProperties>
                 </configuration>
             </plugin>
         </plugins>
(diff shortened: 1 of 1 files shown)


shroffk:
@minijackson hope this addresses your use case

jacomago (approved):

minijackson:
@shroffk It almost works out of the box for me. The issue for me is that since I download GitHub tarballs, the `.git` folder is not present, and the `git-commit-id` Maven plugin fails with:

```
[ERROR] Failed to execute goal io.github.git-commit-id:git-commit-id-maven-plugin:9.0.1:revision (get-the-git-infos) on project ChannelFinder:
    .git directory is not found! Please specify a valid [dotGitDirectory] in your project -> [Help 1]
```

It can easily be fixed on my end, but other users using the GitHub tarball might encounter this error.

jacomago:
@minijackson good point. We should add a default incase the build folder is not a git repository.

shroffk:
@jacomago @minijackson 
https://github.com/git-commit-id/git-commit-id-maven-plugin/blob/8d15df585877c1f13dab2369dd20be91cc45e442/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L568

maybe we should set this to false `failOnNoGitDirectory`

not sure what the default should be for the timestamp ( maybe we should just skip it and just reply on the git.properties included in the jar )

PR #163, opened 2024-10-08
devcontainer added

stan-dot:


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,41 @@
+// For format details, see https://containers.dev/implementors/json_reference/
+{
+    "name": "Java Developer Container",
+    "build": {
+        "dockerfile": "../Dockerfile",
+        "target": "developer"
+    },
+    "remoteEnv": {
+        // Allow X11 apps to run inside the container
+        "DISPLAY": "${localEnv:DISPLAY}"
```
jacomago: ChannelFinder has no desktop application, so I don't think this is necessary.

Dockerfile
```diff
@@ -1,6 +1,27 @@
-FROM eclipse-temurin:17-jre
+# Stage 1: Development stage
+FROM eclipse-temurin:17-jdk AS developer
 
-# deployment unit
+# Install any development tools you need, for example, Maven, Git, etc.
+RUN apt-get update && \
+    apt-get install -y maven git && \
+    apt-get clean
```
jacomago, line 9: Will want docker as well for running the testcontainers based integration tests.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,41 @@
...
+    },
+    "features": {
+        // Some default things like git config
+        "ghcr.io/devcontainers/features/common-utils:2": {
+            "upgradePackages": false
+        }
+    },
+    "runArgs": [
+        // Allow the container to access the host X11 display and EPICS CA
+        "--net=host",
```
jacomago, line 31: Not sure why you would need this? Wouldn't you run any test IOCs in containers?

jacomago (changes requested):

.devcontainer/devcontainer.json
stan-dot: deleted

Dockerfile
stan-dot: hmm, I haven't worked with testcontainers before. I think the docker-in-docker would use the host docker through a socket.

Or rather 'podman in podman' for the DLS users.

.devcontainer/devcontainer.json
stan-dot: I just copy-pasted a generic `devcontainer.json` used in most of our repos. For some developer flows the EPICS CA might be useful

Dockerfile
jacomago: I don't really want to approve without knowing the following command passes in a devcontainer:

`mvn clean install test-compile failsafe:integration-test failsafe:verify --batch-mode --fail-at-end -Djacoco.skip=true -DskipITs=false -DskipITCoverage=true`

stan-dot: will test this

stan-dot: in which instance would you like to run this?

I could drop it here instead of line 23 in Dockerfile

RUN mvn clean install

stan-dot: running this now inside my container.

I am not sure what to expect.

I gest a lot of ConnectionRefused errors.

it expects to connect with Phoebus? not sure

stan-dot: > [ERROR] Tests run: 186, Failures: 6, Errors: 124, Skipped: 0

jacomago: It expects to connect to a locally running container. You might need to set some env variables for this to work from https://java.testcontainers.org/supported_docker_environment/ 

Essentially, there are a bunch of totally end to end tests in that they spin up a container with channel finder, another with postgres, then send http requests to the container to test it. They use the test containers framework for this purpose. 

So to me, if we use a devcontainer it makes sense to have the ability to debug these tests.

stan-dot: of course, I've come to suspect it must be the end to end testcontainers stuff, trying to set this up

Thanks for the pointer :)

stan-dot: I realized that adding testcontainers support is more than this ticket was for originally and it's a vastly new territory as we don't use testcontainers at Diamond while I am quite familiar with the devcontainer setup. 

Theoretically one could have one window open in devcontainer and the other one locally, and run the integration tests there. That would be a quick temporary solution. I tried setting this up and go this error:

The above error is from running locally, after `module load maven` on a DLS device.

> INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  4.306 s
[INFO] Finished at: 2024-10-30T11:12:28Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal io.github.git-commit-id:git-commit-id-maven-plugin:9.0.1:revision (get-the-git-infos) on project ChannelFinder: Execution get-the-git-infos of goal io.github.git-commit-id:git-commit-id-maven-plugin:9.0.1:revision failed: Unable to load the mojo 'revision' in the plugin 'io.github.git-commit-id:git-commit-id-maven-plugin:9.0.1' due to an API incompatibility: org.codehaus.plexus.component.repository.exception.ComponentLookupException: pl/project13/maven/git/GitCommitIdMojo has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
[ERROR] -----------------------------------------------------
[ERROR] realm =    plugin>io.github.git-commit-id:git-commit-id-maven-plugin:9.0.1
[ERROR] strategy = org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy
[ERROR] urls[0] = file:/home/xma12127/.m2/repository/io/github/git-commit-id/git-commit-id-maven-plugin/9.0.1/git-commit-id-maven-plugin-9.0.1.ja

stan-dot: some way of doing this has been demonstrated here https://github.com/gilesknap/docker-outside-docker

with two less updated ways:
- https://confluence.diamond.ac.uk/pages/viewpage.action?spaceKey=~hgv27681&title=Podman+Only
- https://github.com/epics-containers/dev-e7/blob/experiment/docs/podmaninside.md

stan-dot: ok so theoretically the `docker` executable is now available inside the developer container. still elastic search connection refused. investigating

stan-dot:
``` docker run hello_world
docker: Cannot connect to the Docker daemon at unix:///run/user/1225263/podman/podman.sock. Is the docker daemon running?.
See 'docker run --help'.```

Comment (stan-dot) [reply Dockerfile:9]: I installed podman but it refuses to run as root

maybe I need to add something about a new user into devcontainer or Dockerfile

Comment (stan-dot) [reply Dockerfile:9]:
```podman run quay.io/podman/hello
Trying to pull quay.io/podman/hello:latest...
Getting image source signatures
Copying blob 81df7ff16254 done   | 
Copying config 5dd467fce5 done   | 
Writing manifest to image destination
ERRO[0013] Unmounting /var/lib/containers/storage/overlay/6ab302e27a11f20f53d9a2ade9743629b9350af8447449a7acc145ff153eca20/merged: invalid argument 
ERRO[0013] Preparing container 4969dcf74b9acdebf49a1f5004e14b630a68b43570231bead1b8df5ed30bda3f: netavark: setns: IO error: Operation not permitted (os error 1) 
Error: mounting storage for container 4969dcf74b9acdebf49a1f5004e14b630a68b43570231bead1b8df5ed30bda3f: creating overlay mount to /var/lib/containers/storage/overlay/6ab302e27a11f20f53d9a2ade9743629b9350af8447449a7acc145ff153eca20/merged, mount_data="lowerdir=/var/lib/containers/storage/overlay/l/XUJMKUCC4OWTZV3Q5QEPVLUMDN,upperdir=/var/lib/containers/storage/overlay/6ab302e27a11f20f53d9a2ade9743629b9350af8447449a7acc145ff153eca20/diff,workdir=/var/lib/containers/storage/overlay/6ab302e27a11f20f53d9a2ade9743629b9350af8447449a7acc145ff153eca20/work": using mount program /usr/bin/fuse-overlayfs: unknown argument ignored: lazytime
fuse: device not found, try 'modprobe fuse' first
fuse-overlayfs: cannot mount: No such file or directory
: exit status 1
```


I tried parametrizing the remoteUser (comment it out), but it's still as root

stan-dot: even `podman pull` does not quite work as expected 


```go

podman pull docker.elastic.co/elasticsearch/elasticsearch:8.16.1
Trying to pull docker.elastic.co/elasticsearch/elasticsearch:8.16.1...
Getting image source signatures
Copying blob 4ca545ee6d5d done   | 
Copying blob e7894aa63b6b done   | 
Copying blob fc92672f5fb7 done   | 
Copying blob deb9b088b6a9 done   | 
Copying blob d06b22f717ef done   | 
Copying blob 9ab872b2449e done   | 
Copying blob 533fb8518242 done   | 
Copying blob 17672f99d9ac done   | 
Copying blob 7936da0b8295 done   | 
Copying blob f4101ba1f689 done   | 
Error: copying system image from manifest list: writing blob: adding layer with blob "sha256:e7894aa63b6b441f711d543f0bd9f3584e66c7aac167b2f7ceac4bab2a89ca48": processing tar file(potentially insufficient UIDs or GIDs available in user namespace (requested 0:42 for /etc/gshadow): Check /etc/subuid and /etc/subgid if configured locally and run "podman system migrate": lchown /etc/gshadow: invalid argument): exit status 1

```

stan-dot:
okay no merge needed, I can just use it for my local setup

PR #165, opened 2024-10-09
add a maven wrapper

jacomago:


shroffk:
Hmm... should we add/keep some references to the old maven stuff.. like a "do you love maven, then you can do this"

tynanford:
IMO we should keep the README documentation with the normal maven commands unless we decide to implement the maven wrapper for all java maven services (phoebus, alarm tools, cfNameServer, etc)

jacomago:
I updated the README. It was quite out of date... 

I tried to make it tied less to a particular release number or OS etc. And to split up Installation/Running and Development more clearly.

tynanford:
Thanks, the clean up looks pretty good to me.

I wonder if the maven build command should be included in the "Installation" section since we have pretty infrequently releases. On the other hand, the link to the wiki pages covers this so that is fine.

We have been using elastic search 8.2.x for all the services but I assume 8.11.x is what ESS uses and works well? Is it fair to say 8.2.x -> 8.11.x should all work as far as we know?

jacomago:
The integration tests use 8.11 and have been for a while. So I think that's fine.

README.md
```diff
@@ -1,155 +1,134 @@
...
 ```
 
-The above command will start the channelfinder service on port 8080 with the default settings,
-which use embedded ldap server with users and roles defined in the [`cf.ldif`](src/main/resources/cf.ldif) file.
+The default authentication includes an embedded ldap server with users and roles defined in
+the [`cf.ldif`](src/main/resources/cf.ldif) file.
 Note that `cf.ldif` contains **default credentials** and should only be used during testing and evaluation.
 
-#### Verification
+### Verification
```
shroffk, line 89: I had to add the extra curl verification steps due to requests from some EPICS developers...
Do you think there is a better way to incorporate them?

jacomago: I think we should consider the swagger documentation the way of looking for other queries to make. That way we aren't duplicating the api documentation in all of swagger, documentation and the readme.

Swagger in fact gives you the curl commands:
image

shroffk: Ok,
Can we add a note about swagger, including its landing URL in the verification section.

jacomago: Done :)

PR #169, opened 2025-02-21
Fix static resource loading in ChannelFinder WebUI

mgonzal1:
Update relative paths for static resources in index.html to ensure proper loading when accessed under a custom context path. Previously, absolute paths caused 404 errors when the service was deployed with a non-root context path.

Example:
https://services.com/channel-finder/


jacomago:
I have a feeling those test failures are due to a github change in docker engine version :(

PR #179, opened 2025-04-07
Timeout and override post support as property values

conorschofield:
Add timeout seconds and query support override as property value for aa.

When testing LBNL ChannelFinder calls to Archiver Appliance were timing out. I'm adding timeout seconds as an aa property value instead of it being hard coded. The default value is 15 seconds.

Adding an override map to AA_STATUS_ENDPOINT_ONLY_SUPPORT_QUERY_VERSION that is settable as an aa property value. At LBNL our archiver versions are set with our make system and are unconventional 
`{"etl_version":"Archiver Appliance Version d3dd9cc-2022-08-08-als_SNAPSHOT_09-August-2023T17-53-53","mgmt_version":"Archiver Appliance Version d3dd9cc-2022-08-08-als_SNAPSHOT_09-August-2023T17-53-53","retrieval_version":"Archiver Appliance Version d3dd9cc-2022-08-08-als_SNAPSHOT_09-August-2023T17-53-53","engine_version":"Archiver Appliance Version d3dd9cc-2022-08-08-als_SNAPSHOT_09-August-2023T17-53-53"}`

This caused ChannelFinder to use getStatusesFromPVLisBody for Archiver Appliance when our Archiver does not support that. This was causing data buffer limit exception limit on max bytes to buffer. 

By setting query_support_override_map you can set the query support override for each archiver.
Example:
`
aa.query_support_override_map={'arch05': true, 'arch03': true, 'arch-ml': true}`

The override value will default to false if the archiver alias name isn't found in the map.

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
+++ b/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
@@ -197,7 +197,7 @@ public class AAChannelProcessor implements ChannelProcessor {
         if (archivePVS.isEmpty()) {
             return result;
         }
-        List<Map<String, String>> statuses = archiverClient.getStatuses(archivePVS, archiverInfo.url(), archiverInfo.version(), archiverInfo.alias());
+        List<Map<String, String>> statuses = archiverClient.getStatuses(archivePVS, archiverInfo.url(), archiverInfo.alias());
         statuses
                 .forEach(archivePVStatusJsonMap -> {
                     String archiveStatus = archivePVStatusJsonMap.get("status");
--- a/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
+++ b/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
@@ -7,6 +7,7 @@ import org.phoebus.channelfinder.entity.Property;
 import org.phoebus.channelfinder.processors.ChannelProcessor;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.beans.factory.annotation.Autowired;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -51,6 +52,7 @@ public class AAChannelProcessor implements ChannelProcessor {
     @Value("${aa.auto_pause:}")
     private List<String> autoPauseOptions;
 
+    @Autowired
     private final ArchiverClient archiverClient = new ArchiverClient();
 
     @Override
--- a/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
+++ b/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
@@ -197,24 +197,16 @@ public class AAChannelProcessor implements ChannelProcessor {
         if (archivePVS.isEmpty()) {
             return result;
         }
-
-        try {
-            List<Map<String, String>> statuses = archiverClient.getStatuses(archivePVS, archiverInfo.url(), archiverInfo.version(), archiverInfo.alias());
-            statuses
-                    .forEach(archivePVStatusJsonMap -> {
-                        String archiveStatus = archivePVStatusJsonMap.get("status");
-                        String pvName = archivePVStatusJsonMap.get("pvName");
-                        String pvStatus = archivePVS.get(pvName).getPvStatus();
-                        ArchiveAction action = pickArchiveAction(archiveStatus, pvStatus);
-                        result.get(action).add(archivePVS.get(pvName));
-                    });
-            return result;
-
-        } catch (JsonProcessingException e) {
-            // problem collecting policies from AA, so warn and return empty list
-            logger.log(Level.WARNING, () -> "Could not get AA pv Status list: " + e.getMessage());
-            return result;
-        }
+        List<Map<String, String>> statuses = archiverClient.getStatuses(archivePVS, archiverInfo.url(), archiverInfo.version(), archiverInfo.alias());
+        statuses
+                .forEach(archivePVStatusJsonMap -> {
+                    String archiveStatus = archivePVStatusJsonMap.get("status");
+                    String pvName = archivePVStatusJsonMap.get("pvName");
+                    String pvStatus = archivePVS.get(pvName).getPvStatus();
+                    ArchiveAction action = pickArchiveAction(archiveStatus, pvStatus);
+                    result.get(action).add(archivePVS.get(pvName));
+                });
+        return result;
     }
 
     private ArchivePVOptions createArchivePV(
--- a/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
+++ b/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
@@ -199,7 +199,7 @@ public class AAChannelProcessor implements ChannelProcessor {
         }
 
         try {
-            List<Map<String, String>> statuses = archiverClient.getStatuses(archivePVS, archiverInfo.url(), archiverInfo.version());
+            List<Map<String, String>> statuses = archiverClient.getStatuses(archivePVS, archiverInfo.url(), archiverInfo.version(), archiverInfo.alias());
             statuses
                     .forEach(archivePVStatusJsonMap -> {
                         String archiveStatus = archivePVStatusJsonMap.get("status");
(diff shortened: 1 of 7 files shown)


jacomago:
That makes sense, but I wonder about the design. Could you just make AA_STATUS_ENDPOINT_ONLY_SUPPORT_QUERY_VERSION configurable instead? Or is that too much maintenance to keep it up to date?

conorschofield:
That maintenance I think would be too much to keep up to date compared to this solution. Getting the archive version and adding it to ChannelFinder AA_STATUS_ENDPOINT_ONLY_SUPPORT_QUERY_VERSION configurable is less readable and harder to maintain.

Our archiver versions are less meaningful and are not easily associated with the archiver alias. It also might change if our archiver get rebuilt (date at the end could change). Adding `d3dd9cc-2022-08-08-als_SNAPSHOT_09-August-2023T17-53-53` to a property value for query version doesn't tell me much at first glance. I'm not sure if this is referring to our 3 minute or 5 minute archiver. While `aa.query_support_override_map={'arch05': true}` is very clear to me that our 5 minute archiver is choosing to override the query support check.

I will work on fixing the Code Analysis fails.

conorschofield:
The correct this malformed SpEL expression for 
` @Value("#{${aa.query_support_override_map:{'default': false}}}")` seems to be a false positive when using @Value to inject directly into a Map: [sonar source issue](https://community.sonarsource.com/t/java-s6857-false-positive-when-using-value-to-inject-directly-into-a-map/111535/17)

src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
```diff
@@ -51,6 +52,7 @@ public class AAChannelProcessor implements ChannelProcessor {
     @Value("${aa.auto_pause:}")
     private List<String> autoPauseOptions;
 
+    @Autowired
```
jacomago, line 55: Not sure why you added this?

src/main/java/org/phoebus/channelfinder/processors/aa/ArchiverClient.java
```diff
@@ -54,17 +56,24 @@ public class ArchiverClient {
...
+    private Map<String, Boolean> querySupportOverrideMap;
 
     private Stream<List<String>> partitionSet(Set<String> pvSet, int pageSize) {
         List<String> list = new ArrayList<>(pvSet);
         return IntStream.range(0, (list.size() + pageSize - 1) / pageSize)
                 .mapToObj(i -> list.subList(i * pageSize, Math.min(pageSize * (i + 1), list.size())));
     }
 
-    List<Map<String, String>> getStatuses(Map<String, ArchivePVOptions> archivePVS, String archiverURL, String archiverVersion) throws JsonProcessingException {
+    List<Map<String, String>> getStatuses(Map<String, ArchivePVOptions> archivePVS, String archiverURL, String archiverVersion, String archiverAlias) throws JsonProcessingException {
```
jacomago: Looks like you can now drop the JsonProcessingException

src/main/java/org/phoebus/channelfinder/processors/aa/ArchiverClient.java
```diff
@@ -54,17 +56,24 @@ public class ArchiverClient {
...
         Set<String> pvs = archivePVS.keySet();
-        if (AA_STATUS_ENDPOINT_ONLY_SUPPORT_QUERY_VERSION.contains(archiverVersion)) {
+        Boolean querySupportOverride = querySupportOverrideMap.getOrDefault(archiverAlias, false);
+        logger.log(Level.INFO, "Query Support Override Map: {0}", querySupportOverrideMap);
+            
+        if (AA_STATUS_ENDPOINT_ONLY_SUPPORT_QUERY_VERSION.contains(archiverVersion) || Boolean.TRUE.equals(querySupportOverride)) {
```
jacomago: I think we can just drop the AA_STATUS_ENDPOINT_ONLY_SUPPORT_QUERY_VERSION entirely and use the new config you added.

jacomago (changes requested):
Can you also rebase to master. thanks

src/main/java/org/phoebus/channelfinder/processors/aa/ArchiverClient.java
conorschofield: Will this break other versions of channel finder that are still using the AA_STATUS_ENDPOINT_ONLY_SUPPORT_QUERY_VERSION?

tynanford:
Hi @conorschofield , could you also include the new settings in the application.properties file with a comment explaining how they work?

https://github.com/ChannelFinder/ChannelFinderService/blob/master/src/main/resources/application.properties

I guess this new property for query support override should default to false for the "default" archiver in the settings.

Also I'm not sure how common the use case is for multiple archivers with different versions. If we want to simplify the setting it could just be a boolean for all archivers in aa.urls. The map is good, just throwing it out there

`aa.query_support_override=true`

src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
conorschofield: This is needed to set the timeout and query override map for ArchiverClient

src/main/java/org/phoebus/channelfinder/processors/aa/ArchiverClient.java
conorschofield: I've removed it now

src/main/java/org/phoebus/channelfinder/processors/aa/ArchiverClient.java
jacomago: I dont think there is a release, so its fine.

jacomago:
Can you also add an update to https://github.com/ChannelFinder/ChannelFinderService/blob/master/src/site/sphinx/aa_processor.rst for the documentation of this configuration.

conorschofield:
> Can you also rebase to master. thanks

It is rebased to master my timeoutAndOverride branch and master are the same.

src/main/java/org/phoebus/channelfinder/processors/aa/ArchiverClient.java
```diff
@@ -26,26 +28,6 @@
 public class ArchiverClient {
     private static final Logger logger = Logger.getLogger(ArchiverClient.class.getName());
     private static final int STATUS_BATCH_SIZE = 100; // Limit comes from tomcat server maxHttpHeaderSize which by default is a header of size 8k
-    private static final List<String> AA_STATUS_ENDPOINT_ONLY_SUPPORT_QUERY_VERSION = List.of("1.1.0",
```
tynanford, line 30: By removing this version list and defaulting `querySupportOverride` to false, this will cause a change in behavior for users relying on the version check right? They would then need to add the new map property to get the same behavior as before?

tynanford: oh I see the other comment about this...

src/main/java/org/phoebus/channelfinder/processors/aa/ArchiverClient.java
tynanford: So everyone should only use released versions of CF? Sounds like we should create releases more often then :)

src/main/java/org/phoebus/channelfinder/processors/aa/ArchiverClient.java
jacomago: Yes I think we should reverse  if (Boolean.TRUE.equals(querySupportOverride)) { 
The newest verison of the archiver supports both query and post, so we should have the default to use the query and have config for using the post method.

jacomago: @conorschofield Sorry for the back and forth on this. But I think it would be better to swap from

aa.urls={'d':'localhost', 'e':'other.aa', 'f':'foo.aa'}
aa.query_support_override_map = {'d':true}

to 

aa.urls={'d':'localhost', 'e':'other.aa'}
aa.post_support = e, f

And flip the 'if' in ArchiverClient.getStatuses

This would probably be safer (I think I was probably too bullish in wanting people to upgrade AA)

conorschofield: done in [change default mode to query](https://github.com/ChannelFinder/ChannelFinderService/pull/179/commits/a32786190c25aa954db4dbd05e28e8f878bc491c)

jacomago (approved):

PR #180, opened 2025-04-24
4.7.3 release

jacomago:


simon-ess (approved):

anderslindho (approved):
Would be nice if CF's versioning was decoupled from the rest of the phoebus applications and services, so we could follow semantic versioning (ping @shroffk)

shroffk:
Broadly I agree...however,
there are some common modules like utility and data representation which do land up connecting the version numbers
we are going to be updating springboot and JDK versions for CF so that would justify moving to 5.0.0 keeping CF in line with Phoebus.

Maybe we keep the major version synchronized and the minor versions can be updated as required by CF releases

anderslindho:
I see.

> we are going to be updating springboot and JDK versions for CF so that would justify moving to 5.0.0 keeping CF in line with Phoebus.

> Maybe we keep the major version synchronized and the minor versions can be updated as required by CF releases

I think it would be better if we avoid trying to keep versions in sync, and instead increment the major digit only when there are incompatible API changes, minor when we add functionality in a backwards compatible manner, and patch when we make backward compatible bug fixes. That way it will be easier to work with other consumers.

PR #181, opened 2025-05-21
Adds an update interval to the property and tag metrics

jacomago:
They take some resources, so if your are querying a lot of properties this can timeout

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/MetricsService.java
+++ b/src/main/java/org/phoebus/channelfinder/MetricsService.java
@@ -7,6 +7,7 @@ import io.micrometer.core.instrument.Tag;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.PropertySource;
+import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Service;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
@@ -16,6 +17,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 import java.util.Map.Entry;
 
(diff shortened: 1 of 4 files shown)


simon-ess:
This seems like a nice addition. However, if the 5 second interval actually a good one? It seems that switching to re-running this that often might just be hammering the service in a different way.

jacomago:
> This seems like a nice addition. However, if the 5 second interval actually a good one? It seems that switching to re-running this that often might just be hammering the service in a different way.

Its configurable.

anderslindho:
5s default seems a bit overkill to me - is there any use case for that time precision?

simon-ess:
Would it  not be possible to update the atomic count every time a tag or a property is added or removed? Or is that perhaps a fool's errand?

(The point being of course that the number of tags or proprties will not change in the last `$INTERVAL` unless something has actually changed like someone manually adding a tag or some such)

jacomago:
> Would it not be possible to update the atomic count every time a tag or a property is added or removed? Or is that perhaps a fool's errand?
> 
> (The point being of course that the number of tags or proprties will not change in the last `$INTERVAL` unless something has actually changed like someone manually adding a tag or some such)

That is a separate metric, that isn't changed in this PR and it is tracked. CF_PROPERTY_COUNT and CF_TAG_COUNT in MetricsService.

jacomago:
> 5s default seems a bit overkill to me - is there any use case for that time precision?

Updated to 10 seconds

anderslindho:
Prometheus default scrape interval is 1m so we could do that or 30s to save CF resources 🔧 

But otherwise LGTM

simon-ess:
One other question: Does this actually have to be an `AtomicLong`? As near as I see it, we update these values every `$INTERVAL` seconds in a single place. What is the need for concurrency protection?

(I mean, other than if the update itself takes longer than the interval... 😄 )

jacomago:
> One other question: Does this actually have to be an `AtomicLong`? As near as I see it, we update these values every `$INTERVAL` seconds in a single place. What is the need for concurrency protection?
> 
> (I mean, other than if the update itself takes longer than the interval... 😄 )

I'm not sure if necessary when using the GaugeBuilder, but generally using Atomics is suggested in the documentation https://docs.micrometer.io/micrometer/reference/concepts/gauges.html

simon-ess (approved):

PR #182, opened 2025-05-27
Improve swagger documentation with response objects and status-codes

imretoth-ess:
Extended swagger documentation for the REST interface with response objects and status-codes

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/ChannelManager.java
+++ b/src/main/java/org/phoebus/channelfinder/ChannelManager.java
@@ -2,6 +2,11 @@ package org.phoebus.channelfinder;
 
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.Lists;
+import io.swagger.v3.oas.annotations.media.ArraySchema;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.responses.ApiResponses;
 import org.phoebus.channelfinder.AuthorizationService.ROLES;
 import org.phoebus.channelfinder.entity.Channel;
 import org.phoebus.channelfinder.entity.Property;
@@ -72,6 +77,22 @@ public class ChannelManager {
      * @param allRequestParams query parameters
      * @return list of all channels
      */
+    @ApiResponses(
+            value = {
+                    @ApiResponse(
+                            responseCode = "200",
+                            description = "List of all channels",
+                            content = @Content(
+                                    array = @ArraySchema(schema = @Schema(implementation = Channel.class)))),
+                    @ApiResponse(
+                            responseCode = "400",
+                            description = "Invalid request",
+                            content = @Content(schema = @Schema(implementation = ResponseStatusException.class))),
+                    @ApiResponse(
+                            responseCode = "500",
+                            description = "Error while trying to find all channels",
+                            content = @Content(schema = @Schema(implementation = ResponseStatusException.class)))
+            })
     @GetMapping
     public List<Channel> query(@RequestParam MultiValueMap<String, String> allRequestParams) {
         return channelRepository.search(allRequestParams).channels();
@@ -85,18 +106,45 @@ public class ChannelManager {
      * @param allRequestParams query parameters
      * @return SearchResult a count to the total number of matches and the first 10k hits
      */
+    @ApiResponses(
+            value = {
+                    @ApiResponse(
+                            responseCode = "200",
+                            description = "The number of matches for the query, and the first 10k channels",
+                            content = @Content(
+                                    array = @ArraySchema(schema = @Schema(implementation = SearchResult.class)))),
+                    @ApiResponse(
+                            responseCode = "400",
+                            description = "Invalid request - response size exceeded",
+                            content = @Content(schema = @Schema(implementation = ResponseStatusException.class))),
+                    @ApiResponse(
+                            responseCode = "500",
+                            description = "Error while trying to find all channels",
+                            content = @Content(schema = @Schema(implementation = ResponseStatusException.class)))
+            })
     @GetMapping("/combined")
     public SearchResult combinedQuery(@RequestParam MultiValueMap<String, String> allRequestParams) {
         return channelRepository.search(allRequestParams);
     }
 
     /**
-     * GET method for quering the number of matches to a multi-parameter query specifying patterns for tags, property values, and
+     * GET method for querying the number of matches to a multi-parameter query specifying patterns for tags, property values, and
      * channel names to match against.
      *
      * @param allRequestParams query parameters
      * @return a total number of channels that match the query parameters
      */
+    @ApiResponses(
+            value = {
+                    @ApiResponse(
+                            responseCode = "200",
+                            description = "Number of channels that matches the criteria",
+                            content = @Content(schema = @Schema(implementation = Long.class))),
+                    @ApiResponse(
+                            responseCode = "500",
+                            description = "Error while trying to count the result for channel-query",
+                            content = @Content(schema = @Schema(implementation = ResponseStatusException.class)))
+            })
     @GetMapping("/count")
     public long queryCount(@RequestParam MultiValueMap<String, String> allRequestParams) {
         return channelRepository.count(allRequestParams);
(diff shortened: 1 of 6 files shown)


jacomago:
Is it possible to replace all the MultiValueMap with more specific list of query params? Most of them I think are for the search and it should be possible to have query params like:

~size
~name
~total_hits
MultiVAlue
tag

imretoth-ess:
> Is it possible to replace all the MultiValueMap with more specific list of query params? Most of them I think are for the search and it should be possible to have query params like:
> 
> ~size ~name ~total_hits MultiVAlue tag

Well, yes and no.
The MultiValueMap key can end with a '!' character which would mean negation for some of that field when querying. In that case either the '!' character should move to the value part of the MultiValueMap, or introduce an additional boolean flag that would mark the negation per queryParam fields. Either one we choose would change the current syntax - if we remove those fields from being parsed when processing the MultiValueMap.

Another thing I just noticed that if the RequestBody is an Iterable<T> type, it is not showing up on swagger. I think it can not decide what structure it should show/parse, since List, Map,... have different structure.

src/main/java/org/phoebus/channelfinder/ChannelManager.java
```diff
@@ -72,6 +77,22 @@ public class ChannelManager {
      * @param allRequestParams query parameters
      * @return list of all channels
      */
+    @ApiResponses(
+            value = {
+                    @ApiResponse(
+                            responseCode = "200",
+                            description = "List of all channels",
```
anderslindho: The javadoc seems wrong here also - this should just be "List of channels" or "List of matching channels"

src/main/java/org/phoebus/channelfinder/ChannelManager.java
```diff
@@ -72,6 +77,22 @@ public class ChannelManager {
      * @param allRequestParams query parameters
      * @return list of all channels
      */
+    @ApiResponses(
```
anderslindho, line 80: I think we should add `@Operation` annotations here and elsewhere

src/main/java/org/phoebus/channelfinder/ChannelManager.java
```diff
@@ -85,18 +106,45 @@ public List<Channel> query(@RequestParam MultiValueMap<String, String> allReques
      * @param allRequestParams query parameters
      * @return SearchResult a count to the total number of matches and the first 10k hits
      */
+    @ApiResponses(
+            value = {
+                    @ApiResponse(
+                            responseCode = "200",
+                            description = "The number of matches for the query, and the first 10k channels",
+                            content = @Content(
+                                    array = @ArraySchema(schema = @Schema(implementation = SearchResult.class)))),
```
anderslindho, line 115: Hmm, doesn't this return a `SearchResult` object?

src/main/java/org/phoebus/channelfinder/ChannelManager.java
```diff
@@ -85,18 +106,45 @@ public List<Channel> query(@RequestParam MultiValueMap<String, String> allReques
...
      * channel names to match against.
      *
      * @param allRequestParams query parameters
      * @return a total number of channels that match the query parameters
      */
+    @ApiResponses(
+            value = {
+                    @ApiResponse(
+                            responseCode = "200",
+                            description = "Number of channels that matches the criteria",
```
anderslindho: "The number of channels matching the query" 🔧

src/main/java/org/phoebus/channelfinder/ChannelManager.java
```diff
@@ -109,6 +157,17 @@ public long queryCount(@RequestParam MultiValueMap<String, String> allRequestPar
      * @param channelName - channel name to search for
      * @return found channel
      */
+    @ApiResponses(
+            value = {
+                    @ApiResponse(
+                            responseCode = "200",
+                            description = "Channels with the desired name",
```
anderslindho: Not channel**s**; "Channel with the specified name" 🔧

src/main/java/org/phoebus/channelfinder/ChannelManager.java
```diff
@@ -132,6 +191,29 @@ public Channel read(@PathVariable("channelName") String channelName) {
...
+                            responseCode = "200",
+                            description = "The created/replaced channel",
+                            content = @Content(schema = @Schema(implementation = Channel.class))),
+                    @ApiResponse(
+                            responseCode = "400",
+                            description = "Invalid request",
+                            content = @Content(schema = @Schema(implementation = ResponseStatusException.class))),
+                    @ApiResponse(
+                            responseCode = "401",
+                            description = "Authorization problem",
```
anderslindho: "Unauthorized" 🔧

src/main/java/org/phoebus/channelfinder/ChannelManager.java
```diff
@@ -172,6 +254,30 @@ public Channel create(@PathVariable("channelName") String channelName, @RequestB
      * @param channels - XmlChannels to be created
```
anderslindho, line 254: Technically out of scope for your PR but this isn't right - both `XmlChannels` and "to be created" (as it's PUT)

src/main/java/org/phoebus/channelfinder/ChannelManager.java
```diff
@@ -172,6 +254,30 @@ public Channel create(@PathVariable("channelName") String channelName, @RequestB
...
+                            description = "The created/replaced channels",
+                            content = @Content(
+                                    array = @ArraySchema(schema = @Schema(implementation = Channel.class)))),
+                    @ApiResponse(
+                            responseCode = "400",
+                            description = "Invalid request",
+                            content = @Content(schema = @Schema(implementation = ResponseStatusException.class))),
+                    @ApiResponse(
+                            responseCode = "401",
+                            description = "Authorization problem",
```
anderslindho: "Unauthorized" 🔧 

(Applies to all following 401s)

src/main/java/org/phoebus/channelfinder/ChannelManager.java
```diff
@@ -351,6 +503,28 @@ public Iterable<Channel> update(@RequestBody Iterable<Channel> channels) {
...
      */
+    @ApiResponses(
+            value = {
+                    @ApiResponse(
+                            responseCode = "200",
+                            description = "Channel deleted"),
+                    @ApiResponse(
+                            responseCode = "400",
+                            description = "Invalid request",
+                            content = @Content(schema = @Schema(implementation = ResponseStatusException.class))),
```
anderslindho: Possibly redundant - is there validation?

src/main/java/org/phoebus/channelfinder/ChannelScroll.java
```diff
@@ -58,6 +62,17 @@ public class ChannelScroll {
      * @param allRequestParams search parameters
      * @return list of all channels
      */
+    @ApiResponses(
+            value = {
+                    @ApiResponse(
+                            responseCode = "200",
+                            description = "List of channels",
```
anderslindho: Doesn't technically return all channels, but rather a paginated scroll result

anderslindho (commented):
Largely looks good to me, but some nitpicks here and there.

@jacomago I don't think we will want to create a release with annotating the operations... Should we create follow-up or handle here? It might 2x the PR and work.

src/main/java/org/phoebus/channelfinder/ChannelManager.java
imretoth-ess: Yes, it responds with a SearchResult object that contains only 2 fields: a count, and a List of Channels - but on the response List length there is a restriction

src/main/java/org/phoebus/channelfinder/ChannelManager.java
imretoth-ess: Yes, you are right, that is not needed here, sorry

src/main/java/org/phoebus/channelfinder/ChannelManager.java
anderslindho: @jacomago thoughts?

src/main/java/org/phoebus/channelfinder/ChannelManager.java
anderslindho: You can ignore this part @imretoth-ess - I will create a sub-task and I can take that myself so you don't have to modify the scope in the middle of the PR

anderslindho:
LGTM

PR #183, opened 2025-06-05
Improve swagger documentation with operation annotations

anderslindho:


src/main/java/org/phoebus/channelfinder/processors/ChannelProcessorManager.java
```diff
@@ -108,6 +127,12 @@ public long processAllChannels() {
         }
     }
 
+    @Operation(
+        summary = "Process channels by query",
+        description = "Manually trigger processing on channels matching the given query.",
```
imretoth-ess, line 129: Should we add here what kind of keywords can be used in the query? Like:
You can use the following keywords as search parameters to filter on specific fields:
            - ~name
            - ~tag
            - ~size
            - ~from
            
            You can use the '!' character at the end of keyword to negate the filtering expression
            (for used keywords only tag can have negating). e.g.: ~tag!
            
            You can use the following characters as dividers for searching for values: [|,;]

anderslindho: I guess that should be done through `@Parameter` annotations, but here only a single one for all params (due to `MultiValueMap<String, String>`)?

jacomago: Yes, Imre pointed out in the last pr that unfortunately changing the multivaluemap would break the API. So for now the best way is to add the information in the operation or the parameter description

anderslindho: Yeah I see that too. Maybe we should have a backlog ticket to at least record this - I assume we will want to change it for a v2 API?

anderslindho: I guess the param annotation really should be in the query methods in ChannelManager and ChannelScroll instead of on the processor, or?

jacomago: > Yeah I see that too. Maybe we should have a backlog ticket to at least record this - I assume we will want to change it for a v2 API?

I think so yes, but v2 API I think we should try design from scratch

> I guess the param annotation really should be in the query methods in ChannelManager and ChannelScroll instead of on the processor, or?

Also here on the processor I think

anderslindho: Turns out it was relevant in even more places, so I extracted the message into a var. Harms readability a little bit, but at least the descriptions should not diverge now.

jacomago (approved):

PR #184, opened 2025-08-19
add extra logging and skipping over bad pvs

conorschofield:
channelfinder would break with a null runtime exception here. I've added logging and allowed channelfinder to skip over bad pvs instead of breaking.

For example, we were seeing problems populating a new archiver with channel finder and couldn't figure out why it wouldn't work with a specific ioc (it would break with null runtime exception). I added this and figured out that channelfinder was having problems archiving pvs with commas. It looks like it is splitting the pvs with commas causing channelfinder to break when getting the actions for those pvs.

Overall, functionality to fix pvs with commas still needs to be fixed but for now this lets channel finder catch bad pvs and skip over them instead of breaking.

Example:
BTS_____Q6,1___BM02:AtInjection
`Aug 19 10:39:13 controls-hlc1.als.lbl.gov java[2889]: 2025-08-19 10:39:13.474  WARN 2889 --- [taskExecutor-89] o.p.c.processors.aa.AAChannelProcessor   : archivePVS does not contain pvName: BTS_____Q6
Aug 19 10:39:13 controls-hlc1.als.lbl.gov java[2889]: 2025-08-19 10:39:13.474  WARN 2889 --- [taskExecutor-89] o.p.c.processors.aa.AAChannelProcessor   : archivePVS does not contain pvName: 1___BM02:AtInjection`

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
+++ b/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
@@ -230,11 +230,7 @@ public class AAChannelProcessor implements ChannelProcessor {
       return result;
     }
     List<Map<String, String>> statuses = archiverClient.getStatuses(archivePVS, archiverInfo.url(), archiverInfo.alias());
-    if (statuses == null) {
-      logger.log(Level.WARNING, "archiverClient.getStatuses returned null");
-      return result;
-    }
-    logger.log(Level.INFO, "Statuses {0}", statuses);
+    logger.log(Level.FINER, "Statuses {0}", statuses);
     statuses.forEach(archivePVStatusJsonMap -> {
       String archiveStatus = archivePVStatusJsonMap.get("status");
       String pvName = archivePVStatusJsonMap.get("pvName");
--- a/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
+++ b/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
@@ -250,10 +250,6 @@ public class AAChannelProcessor implements ChannelProcessor {
       ArchiveAction action = pickArchiveAction(archiveStatus, pvStatus);
 
       List<ArchivePVOptions> archivePVOptionsList = result.get(action);
-      if (archivePVOptionsList == null) {
-        logger.log(Level.WARNING, "No list found for action: {0}", action);
-        return;
-      }
       archivePVOptionsList.add(archivePVOptions);
     });
     return result;
--- a/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
+++ b/src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
@@ -198,14 +198,36 @@ public class AAChannelProcessor implements ChannelProcessor {
             return result;
         }
         List<Map<String, String>> statuses = archiverClient.getStatuses(archivePVS, archiverInfo.url(), archiverInfo.alias());
-        statuses
-                .forEach(archivePVStatusJsonMap -> {
-                    String archiveStatus = archivePVStatusJsonMap.get("status");
-                    String pvName = archivePVStatusJsonMap.get("pvName");
-                    String pvStatus = archivePVS.get(pvName).getPvStatus();
-                    ArchiveAction action = pickArchiveAction(archiveStatus, pvStatus);
-                    result.get(action).add(archivePVS.get(pvName));
-                });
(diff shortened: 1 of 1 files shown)


jacomago:
Your naming convention allows PVs with commas! 😥

Does the archiver actually handle that? Both archiving and retrieval of data? Especially the getDataAtTime api?

I would really recommend URL safe pv names if you can.

conorschofield:
> Your naming convention allows PVs with commas! 😥
> 
> Does the archiver actually handle that? Both archiving and retrieval of data? Especially the getDataAtTime api?
> 
> I would really recommend URL safe pv names if you can.

Unfortunately we have some iocs that have pvs with commas (legacy iocs)

To my knowledge some of these pvs with commas are working with the archiver. It could be they are working because of aliases but I am not sure.

src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
```diff
@@ -198,14 +198,36 @@ private Map<ArchiveAction, List<ArchivePVOptions>> getArchiveActions(
...
+                    logger.log(Level.WARNING, "archivePVS does not contain pvName: {0}", pvName);
+                    return;
+                }
+
+                String pvStatus = archivePVOptions.getPvStatus();
+                ArchiveAction action = pickArchiveAction(archiveStatus, pvStatus);
+
+                List<ArchivePVOptions> archivePVOptionsList = result.get(action);
+                if (archivePVOptionsList == null) {
+                    logger.log(Level.WARNING, "No list found for action: {0}", action);
```
jacomago: I'm not sure its possible for archivePVOptionsList to be null, see lines 194 to 195.

src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
```diff
@@ -198,14 +198,36 @@ private Map<ArchiveAction, List<ArchivePVOptions>> getArchiveActions(
...
-                    String pvName = archivePVStatusJsonMap.get("pvName");
-                    String pvStatus = archivePVS.get(pvName).getPvStatus();
-                    ArchiveAction action = pickArchiveAction(archiveStatus, pvStatus);
-                    result.get(action).add(archivePVS.get(pvName));
-                });
+        if (statuses == null) {
+            logger.log(Level.WARNING, "archiverClient.getStatuses returned null");
+            return result;
+        }
+        logger.log(Level.INFO, "Statuses {0}", statuses);
```
jacomago: I think this should be debug level (or FINER in java logging levels). This could be massive.

src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
```diff
@@ -198,14 +198,36 @@ private Map<ArchiveAction, List<ArchivePVOptions>> getArchiveActions(
...
-        statuses
-                .forEach(archivePVStatusJsonMap -> {
-                    String archiveStatus = archivePVStatusJsonMap.get("status");
-                    String pvName = archivePVStatusJsonMap.get("pvName");
-                    String pvStatus = archivePVS.get(pvName).getPvStatus();
-                    ArchiveAction action = pickArchiveAction(archiveStatus, pvStatus);
-                    result.get(action).add(archivePVS.get(pvName));
-                });
+        if (statuses == null) {
+            logger.log(Level.WARNING, "archiverClient.getStatuses returned null");
```
jacomago: This should be impossible. If you see archiverClient.getStatusesFromPvListQuery and archiverClient.getStatusesFromPvListBody return empty lists rather than null if something goes wrong. Or they should do.

jacomago:
Have you tried with an archiver > 2.0.9 setting aa.post_support to them. I think posting the list of statuses rather than using a query should work?

src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
conorschofield: I've removed this in the latest commit

src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
conorschofield: changed to FINER

src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
conorschofield: removed this check

jacomago (approved):

Issue #186, opened 2025-10-06
Errors when indexing large # of channels


shroffk:
A recent installation of ChannelFinder encountered an issue when a batch index/save of a140k channels breached the elastic search 100mb request limit.

While the immediate solution would be to configure the 
`http.max_content_length: 100mb`

However, it might be nice to have the service break up very large request too.

PR #187, opened 2025-10-07
Chunking

shroffk:
Break very large index and update request into smaller chunks
Address the issue #186

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/ChannelRepository.java
+++ b/src/main/java/org/phoebus/channelfinder/ChannelRepository.java
@@ -173,7 +173,7 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
     List<Channel> allIndexed = new ArrayList<>();
     for (Future<List<Channel>> future : futures) {
       try {
-        allIndexed.addAll(future.get());
+        allIndexed.addAll(future.get(10, TimeUnit.MINUTES));
       } catch (Exception e) {
         logger.log(Level.SEVERE, "Bulk indexing failed", e);
       }
--- a/src/main/java/org/phoebus/channelfinder/ChannelRepository.java
+++ b/src/main/java/org/phoebus/channelfinder/ChannelRepository.java
@@ -31,22 +31,29 @@ import co.elastic.clients.json.jackson.JacksonJsonpMapper;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
 import java.text.MessageFormat;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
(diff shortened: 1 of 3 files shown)


jacomago:
Oh, you're ahead of me https://github.com/ChannelFinder/ChannelFinderService/pull/188

shroffk:
An anomaly for sure...
I did make the process parallel too, it might be overkill.

jacomago:
> An anomaly for sure... I did make the process parallel too, it might be overkill.

Yes, not sure about the parallelization. Nor using skip. 

I think some combination of the two would be good. I forgot to use @Value for instance

shroffk:
Well I was trying to find the simplest way to not make copies and also support multi threading
What is your primary concern with the skip

jacomago:
> Well I was trying to find the simplest way to not make copies and also support multi threading What is your primary concern with the skip

I just think then you have the extra loop. Wheras mine just does on loop still. But I'm not sure if it matters much.

shroffk:
well I am hoping that the multi threaded bit means that when you have large number of channels being indexed 
( like populating for performance testing, or like in ESS's case where the whole CF in wiped out an recreated ) then doing it in chunks sequentially might have some performance issues.

shroffk:
@tynanford do you want to share your opinion

tynanford:
tagging @conorschofield  as well

should `index.chunk.size` and `processors.chunking.size` have the same default of 10K?

I don't know enough java to have an opinion on skip or how to structure the loops. We also do the same as ESS and re-populate the entire CF instance every so often. All the CF data is stored in IOCs or in a matlab script which adds MML meta-data to CF . So parallelization sounds good to me.

jacomago:
I'm wondering if some of the tests are breaking because there is no longer a consistent ordering of the returned channels.

jacomago:
> should `index.chunk.size` and `processors.chunking.size` have the same default of 10K?
> 

I based it on the default elastic window size. 10 000 seems to be fine most of the time anyway. We hit the limit at a 170 000 IOC, I calculated that 110 000 is around where the problem is so I think 10 000 is a good default.

shroffk:
> I'm wondering if some of the tests are breaking because there is no longer a consistent ordering of the returned channels.

I don't think that is the case with the manual IT tests.

Maybe we can have one preference for both/all the chunking operations.

src/main/java/org/phoebus/channelfinder/ChannelRepository.java
```diff
@@ -110,41 +124,60 @@ public Channel index(Channel channel) {
...
-            logger.log(Level.SEVERE, () -> item.error().reason());
-          }
-        }
-        // TODO cleanup? or throw exception?
-      } else {
-        return channels;
+    List<Channel> allIndexed = new ArrayList<>();
+    for (Future<List<Channel>> future : futures) {
+      try {
+        allIndexed.addAll(future.get());
```
georgweiss: Does the ```get()``` call need a timeout to avoid waiting forever?

src/main/java/org/phoebus/channelfinder/ChannelRepository.java
```diff
@@ -67,11 +76,18 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
   @Qualifier("indexClient")
   ElasticsearchClient client;
 
+  @Value("${index.chunk.size:1}")
```
tynanford: I would either have one setting for the chunk size for all operations or make the defaults the same for both (10K sounds good to me). If keeping `index.chunk.size` then we should add it to the application.properties file as well for visibility

PR #190, opened 2025-10-13
Autoformatter

jacomago:


jacomago:
Got fed up with the lack of auto formatted code.

shroffk (approved):

PR #192, opened 2025-10-28
Make processor enabled runtime

jacomago:


shroffk:
Do we have to remove the preference based switch
I see the benefit in having a end point to enable/disable this processor but it is also nice to be able to manage this using preferences in a more static setup

jacomago:
Hmmm, yes

> Do we have to remove the preference based switch I see the benefit in having a end point to enable/disable this processor but it is also nice to be able to manage this using preferences in a more static setup

Hmmm, yes. I wonder what to do about reboots. I'll set it so the properties manages the persisted value at least.

src/main/java/org/phoebus/channelfinder/processors/ChannelProcessorManager.java
```diff
@@ -151,4 +154,9 @@ public long processChannels(
   public void processChannels(List<Channel> channels) {
     channelProcessorService.sendToProcessors(channels);
   }
+
+  @PutMapping("/processor/enableProcessor")
```
anderslindho: Should we be extending with non-RESTful endpoints?

src/main/java/org/phoebus/channelfinder/processors/ChannelProcessorManager.java
```diff
@@ -75,10 +75,13 @@ public long processorCount() {
...
             description = "List of processor-info",
             content =
-                @Content(array = @ArraySchema(schema = @Schema(implementation = String.class))))
+                @Content(
+                    array =
+                        @ArraySchema(
+                            schema = @Schema(implementation = ChannelProcessorInfo.class))))
       })
-  @GetMapping("/info")
-  public List<String> processorInfo() {
```
anderslindho, line 81: Should this first be deprecated, or no need to care?

src/main/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessor.java
```diff
@@ -58,7 +58,7 @@ public class AAChannelProcessor implements ChannelProcessor {
 
   private boolean aaEnabled;
 
-  public AAChannelProcessor(@Value("${aa.enabled:true}")  boolean aaEnabled) {
+  public AAChannelProcessor(@Value("${aa.enabled:true}") boolean aaEnabled) {
```
anderslindho: Didn't you already auto-format every file in this project?

georgweiss (commented):
Are the new endpoints covered by unit tests? Asking as I usually find it useful to add unit tests to be able to capture minor changes, e.g. change name or method of endpoint.

src/main/java/org/phoebus/channelfinder/processors/ChannelProcessorManager.java
jacomago: I don't think care.

src/main/java/org/phoebus/channelfinder/processors/ChannelProcessorManager.java
jacomago: Better?

anderslindho: Yup 👍

jacomago:
> Are the new endpoints covered by unit tests? Asking as I usually find it useful to add unit tests to be able to capture minor changes, e.g. change name or method of endpoint.

Yes see https://github.com/ChannelFinder/ChannelFinderService/pull/192/commits/d471b19b4f932dd7f99dc74f6889efaf1e04c5ef

jacomago:
> Are the new endpoints covered by unit tests? Asking as I usually find it useful to add unit tests to be able to capture minor changes, e.g. change name or method of endpoint.

Added a test for the Manager (equivalent to resource in olog)

src/test/java/org/phoebus/channelfinder/processors/ChannelProcessorServiceTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @BeforeEach
+  void setUp() {
+    channelProcessorService =
+        new ChannelProcessorService(List.of(dummyProcessor), Runnable::run, 10);
+  }
+
+  @Test
+  void testEnableAndDisableDummyProcessor() {
+    // Initially, the dummy processor should be disabled
```
anderslindho: Not really any value added from these comments when ~ the same messages are in the assert messages

anderslindho (approved):
LGTM

PR #194, opened 2025-12-05
Upgrade Spring Boot and Java

domonkos-ess:
Changes:

- Upgrade Spring Boot version to 3.4.4
- Upgrade Java version to 21
- Upgrade Jackson version to 2.18.3
- Upgrade Elasticsearch version to 8.18.0
- Align Docker related files
- Refactor WebSecurityConfig accordingly


georgweiss (commented):
For the Olog service we're about to switch to:
* Java 25 (to benefit from language features like lightweight threads)
* Spring Boot 3.5.7
* Jackson 3.0.1
Would this be viable for Channel Finder aswell?

anderslindho:
@georgweiss we specified these versions for now, but we can create follow up for java 25 + highest spring boot 3, and then follow that with spring boot 4

README.md
```diff
@@ -98,7 +98,7 @@ and [Eclipse](https://eclipseide.org/).
 
 * Prerequisites
 
-    * JDK 17
+    * JDK 21
     * Maven (via package manager or via the wrapper `./mvnw`) (version specified
       in [the wrapper properties](./.mvn/wrapper/maven-wrapper.properties))
```
anderslindho, line 103: Don't you need to update also this? (maven wrapper)

anderslindho (dismissed):
LGTM

@jacomago do we wanna set a requirement for the sonar lints - whether they must be addressed?

and do we want to do something about the tests?

georgweiss:
@anderslindho, since the Channel Finder service is an integral part of the Phoebus stack, it might be seen as a bit confusing to not run on the same base dependencies.
What is a "follow up" here in terms of planing and time schedule?

README.md
domonkos-ess: Should I upgrade it to 3.9+?

anderslindho:
@georgweiss

We do not have a precise time plan, as we are agile.

We can bump this in prio based on value, if you think this can cause confusion for maintainers. It would, however, be categorised as a maintenance change, to be prioritised lower than support and dev. Regardless, we often manage to squeeze in smaller maintenance updates in between support and dev work.

Do you have any targets for the rest of the services?

compose.yml
```diff
@@ -15,7 +15,7 @@ services:
         condition: service_healthy
 
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:8.11.2
+    image: docker.elastic.co/elasticsearch/elasticsearch:8.18.0
```
jacomago: Can we update elastic in a different MR? It changes the requirements for running.

src/main/java/org/phoebus/channelfinder/configuration/WebSecurityConfig.java
```diff
@@ -83,71 +101,113 @@ public void configure(WebSecurity web) throws Exception {
...
-    }
+    };
   }
 
   @Bean
   public PasswordEncoder encoder() {
     return new BCryptPasswordEncoder();
   }
+
+  private static class EmbeddedLdapCondition implements Condition {
```
jacomago: This seems mixed up with demo auth and embedded ldap. They should be separate beans.

src/test/java/org/phoebus/channelfinder/processors/ChannelProcessorControllerIT.java
```diff
@@ -30,7 +30,7 @@
 class ChannelProcessorManagerIT {
 
   protected static final String AUTHORIZATION =
-      "Basic " + Base64Utils.encodeToString("admin:adminPass".getBytes());
+      "Basic " + Base64.getEncoder().encodeToString("admin:adminPass".getBytes());
```
jacomago, line 34: Getting failures on this test. I'm guessing because of the mix of embeddedldap and demo auth.

jacomago (changes requested):

georgweiss:
@anderslindho, Olog will be merged this week. For the other services I expect these to be migrated early next year.

README.md
anderslindho: Let's keep it out of scope - please create a backlog task re maven version

src/main/java/org/phoebus/channelfinder/configuration/WebSecurityConfig.java
jacomago: Can you rename this to DemoAuthCondition

anderslindho:
Did you check on the sonar issue @domonkos-ess ?

domonkos-ess:
> Did you check on the sonar issue @domonkos-ess ?

Yes, I saw it, however this is the equivalent of the previous implementation. SonarQube just didn’t take that into account. I documented the code.

jacomago (approved):

PR #195, opened 2026-02-09
Update java to java 25

simon-ess:
Updates to java 25 as well as a few other package updates, and minor update-related cleanup.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,7 +42,9 @@
         <skipITs>true</skipITs>
         <skipITCoverage>true</skipITCoverage>
         <jacoco.skip>true</jacoco.skip>
-        <java.version>21</java.version>
+        <java.version>25</java.version>
+        <maven.compiler.source>25</maven.compiler.source>
+        <maven.compiler.target>25</maven.compiler.target>
         <!--suppress UnresolvedMavenProperty -->
         <project.build.outputTimestamp>${git.commit.time}</project.build.outputTimestamp>
     </properties>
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -35,14 +35,16 @@
     </developers>
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <spring.boot-version>3.4.4</spring.boot-version>
-        <jackson.version>2.18.3</jackson.version>
-        <elasticsearch.version>8.11.2</elasticsearch.version>
+        <spring.boot-version>3.5.7</spring.boot-version>
+        <jackson.version>3.0.1</jackson.version>
+        <elasticsearch.version>8.18.0</elasticsearch.version>
```
anderslindho: Is this (elastic) really required?

anderslindho (commented):
Largely looks fine, but you missed README - should maybe grep for further potential mentions of java 21 (also for elastic if you keep that upgrade). You should, however, inspect failing CI - looks like you might be missing some dep/plugin update (`Failed to execute goal org.jacoco:jacoco-maven-plugin:0.8.10:report (report)`); are you maybe missing an update of Maven? It is used at least for CI. If so, check also for maven wrapper, as that might need an update too.

A few nitpicks:
- 4 new sonar issues introduced, 2 of which seem to relate to the java version (it doesn't like you naming the exception `ex` for some reason)
- would maybe be better to do dep update before java25 update, to have each commit pass build, etc.

jacomago (dismissed):

simon-ess:
Code coverage is failing, but it looks like a docker API thing.

anderslindho:
Don't we need to update maven (wrapper)?

simon-ess:
What needs to be updated in the wrapper? I see no reference to jdk versions in there.

anderslindho:
IIRC I had to update it for CE, but don't remember why...

simon-ess:
It seems like it is not necessary, but I may as well, why not...

jacomago (approved):

PR #197, opened 2026-02-11
Remove ignorable error logs

jacomago:


src/test/java/org/phoebus/channelfinder/MetricsServiceTest.java
```diff
@@ -67,7 +67,7 @@ void testGenerateAllMultiValueMaps_emptyMap() {
     List<MultiValueMap<String, String>> allMaps =
         MetricsService.generateAllMultiValueMaps(properties);
 
-    assertEquals(1, allMaps.size());
+    assertEquals(0, allMaps.size());
```
simon-ess, line 70: Huh. Explain?

jacomago: Don't know why I had that before, but since I updated MetricsService to now return an empty map instead of a list with an emtpymap, now this should be 0.

anderslindho (approved):
You have to put more info in your commit messages: it doesn't make sense that I have to read and understand the code in order to understand the commit message and then, following that, what you're wanting to accomplish and why.

PR #199, opened 2026-02-13
Update elastic in pom.xml

simon-ess:


simon-ess:
Note that in the previous PR (#195 ) I had actually updated it everywhere _else_ but the pom file.

jacomago (approved):

anderslindho (approved):

PR #207, opened 2026-04-02
Refactor: Introduce service layer and thin down controller layer

anderslindho:
Moves towards a clean controller -> service -> repository layering as the foundation for a future versioned REST API.


src/main/java/org/phoebus/channelfinder/service/external/ArchiverService.java
```diff
@@ -61,11 +61,10 @@ String key() {
   private List<String> postSupportArchivers;
 
   @Autowired
-  public ArchiverService(
-      RestClient.Builder builder, @Value("${aa.timeout_seconds:15}") int timeoutSeconds) {
+  public ArchiverService(@Value("${aa.timeout_seconds:15}") int timeoutSeconds) {
     SimpleClientHttpRequestFactory factory = new SimpleClientHttpRequestFactory();
     factory.setReadTimeout(timeoutSeconds * 1000);
-    this.client = builder.requestFactory(factory).build();
+    this.client = RestClient.builder().requestFactory(factory).build();
```
jacomago: The builder was autowired on purpose so it can inherit the configuration from the spring config in the properties file before being configured here.

src/main/java/org/phoebus/channelfinder/service/InfoService.java
```diff
@@ -0,0 +1,58 @@
...
+public class InfoService {
+
+  private static final Logger logger = Logger.getLogger(InfoService.class.getName());
+
+  private static final ObjectMapper objectMapper =
+      new ObjectMapper().enable(SerializationFeature.INDENT_OUTPUT);
+
+  private final ElasticConfig esService;
+
+  @Value("${channelfinder.version:4.7.0}")
```
jacomago: this version number is out of date.

src/main/java/org/phoebus/channelfinder/configuration/WebSecurityConfig.java
```diff
@@ -27,22 +27,40 @@
...
         .authorizeHttpRequests(auth -> auth.anyRequest().authenticated())
         .httpBasic(withDefaults())
         .build();
   }
 
+  @Bean
+  public CorsConfigurationSource corsConfigurationSource() {
+    CorsConfiguration config = new CorsConfiguration();
+    config.setAllowedOriginPatterns(corsAllowedOrigins);
+    config.setAllowedMethods(List.of("GET", "POST", "PUT", "DELETE", "PATCH", "OPTIONS"));
```
jacomago: Surely there must be a way of getting this from a spring boot enum

jacomago (commented):
I don't particularly like the name legacy, I would prefer v0 or something clearer. Legacy is confusing as if you have two things called legacy or old, you don't know what the sequence is.

src/main/java/org/phoebus/channelfinder/web/v0/V0ExceptionHandler.java
imretoth-ess, line 1: Personally I would add a default handler if an exception is being thrown that is not listed here (Catches Exception.class + logs message + returns a default HTTP status code)

src/main/java/org/phoebus/channelfinder/configuration/WebSecurityConfig.java
anderslindho: Yes and no. We can get them but we'll still need to handpick since we otherwise also get HEAD and TRACE: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpMethod.html

src/main/java/org/phoebus/channelfinder/service/InfoService.java
anderslindho: Not really my scope but I can fix

jacomago:
Weird you manage to have both code coverage and integration tests working and now they are failing again...

anderslindho:
> Weird you manage to have both code coverage and integration tests working and now they are failing again...

The fall-back exception handler 🙃 

FWIW I always had either both passing or both failing, in case you are insinuating that the tests historically have been flaky.

imretoth-ess (dismissed):
LGTM! 👍

jacomago (dismissed):
I would prefer anything you added called legacy be v0 instead.

jacomago (approved):

PR #209, opened 2026-04-14
Spring Boot 4

anderslindho:


src/main/java/org/phoebus/channelfinder/configuration/HttpConnectorConfig.java
```diff
@@ -3,28 +3,21 @@
...
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.context.annotation.PropertySource;
 
 @Configuration
-@PropertySource(value = "classpath:application.properties")
 public class HttpConnectorConfig {
 
-  @Value("${server.http.enable:true}")
-  private boolean httpEnabled;
```
simon-ess, line 17: Should this be removed from application.properties?

https://github.com/ChannelFinder/ChannelFinderService/blob/master/src/main/resources/application.properties#L6

src/main/java/org/phoebus/channelfinder/service/InfoService.java
```diff
@@ -11,14 +8,18 @@
...
+import tools.jackson.databind.json.JsonMapper;
 
 @Service
 public class InfoService {
 
   private static final Logger logger = Logger.getLogger(InfoService.class.getName());
 
   private static final ObjectMapper objectMapper =
-      new ObjectMapper().enable(SerializationFeature.INDENT_OUTPUT);
+      JsonMapper.builder().enable(SerializationFeature.INDENT_OUTPUT).build();
```
simon-ess, line 23: I'm a bit surprised at this change, but looking at the docs this seems correct (https://www.javadoc.io/doc/tools.jackson.core/jackson-databind/3.0.0-rc4/tools.jackson.databind/tools/jackson/databind/json/JsonMapper.html)

src/test/java/org/phoebus/channelfinder/docker/ChannelFinderChannelsIT.java
```diff
@@ -33,6 +32,7 @@
 import org.testcontainers.containers.ComposeContainer;
 import org.testcontainers.junit.jupiter.Container;
 import org.testcontainers.junit.jupiter.Testcontainers;
+import tools.jackson.databind.ObjectMapper;
```
simon-ess, line 35: Does this belong in the previous commit?

src/test/java/org/phoebus/channelfinder/docker/ChannelFinderPropertiesIT.java
```diff
@@ -34,6 +33,7 @@
 import org.testcontainers.containers.ComposeContainer;
 import org.testcontainers.junit.jupiter.Container;
 import org.testcontainers.junit.jupiter.Testcontainers;
+import tools.jackson.databind.ObjectMapper;
```
simon-ess, line 36: Same here and the next few changes. They all look like Jackson migration which doesn't match the commit message.

src/main/java/org/phoebus/channelfinder/configuration/HttpConnectorConfig.java
anderslindho: I think it must be kept - it is still consumed by @ConditionalOnProperty on the HTTP connector customizer.

imretoth-ess (approved):
LGTM

src/main/java/org/phoebus/channelfinder/configuration/AAChannelProcessor.java
```diff
@@ -98,10 +98,10 @@ public ChannelProcessorInfo processorInfo() {
    *
    * @param channels List of channels
    * @return Return number of channels processed
-   * @throws JsonProcessingException If processing archiver responses fail.
+   * @throws JacksonException If processing archiver responses fail.
```
jacomago, line 101: Can be a follow up, but maybe we should have an internal JsonParsingException.

jacomago (approved):

src/main/java/org/phoebus/channelfinder/configuration/AAChannelProcessor.java
anderslindho: Ticket created

Issue #211, opened 2026-05-01
Expand query/search API to support query body for bulk channel retrieval


shroffk:
## Context
While reviewing the recceiver/cfstore implementation, inefficiencies in the ChannelFinder Service API were identified, especially when handling bulk channel queries.

### Current Limitation
The current API supports name queries like:
```
/ChannelFinder/resources/channels?~name=somePV1,otherPV2,anotherPV3,...
```
However, this approach is limited by the maximum URL length (~600 characters), making it cumbersome and inefficient to retrieve a large number of channels whose names cannot be easily combined using a single regex. This leads to a significant increase in the number of REST calls required for bulk lookups.

## Suggestion
To address this, follow the approach used by Olog and Elastic by supporting a query body for search operations. This would enable bulk searches and more advanced queries, avoiding the URL length limitation and optimizing performance for clients needing channel data in batches.

### Proposal
- Extend the search/query API to accept a request body (e.g., JSON array of channel names or filters).
- Maintain backward compatibility with existing query parameters.

PR #212, opened 2026-05-01
Query body

shroffk:
#211 

notes: 

1.
```
 // Backward-compatible overload when no request body is provided.
  default Scroll query(String scrollId, MultiValueMap<String, String> searchParameters) {
    return query(scrollId, searchParameters, null);
  }
```

this bit is primarily for the Tests and not used for anything else

2.
Currently, I try to merge the query parameters from the URL and the request body. This is good for backward compatibility but we could simplify this by only allowing one or the other or simply prioritize one or the other.

Neither of the above points are show stoppers and I think this addition would simplify and improve performance of clients.

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/web/v0/api/IChannel.java
+++ b/src/main/java/org/phoebus/channelfinder/web/v0/api/IChannel.java
@@ -10,6 +10,7 @@ import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.responses.ApiResponses;
 import java.util.List;
+import java.util.Map;
 import org.phoebus.channelfinder.entity.Channel;
 import org.phoebus.channelfinder.entity.SearchResult;
 import org.springframework.util.MultiValueMap;
@@ -27,7 +28,10 @@ public interface IChannel {
   @Operation(
       summary = "Query channels",
       description =
-          "Query a collection of Channel instances based on tags, property values, and channel names.",
+          "Query a collection of Channel instances based on tags, property values, and channel names. "
+              + "Search parameters can be provided via URL query string or JSON request body, or both. "
+              + "URL parameters take precedence for control parameters (~size, ~from, ~search_after, ~track_total_hits). "
+              + "Regular search parameters from URL and body are combined as separate values in the query.",
       operationId = "queryChannels",
       tags = {"Channel"})
   @ApiResponses(
@@ -49,12 +53,25 @@ public interface IChannel {
   @GetMapping
   List<Channel> query(
       @Parameter(description = SEARCH_PARAM_DESCRIPTION) @RequestParam
-          MultiValueMap<String, String> allRequestParams);
+          MultiValueMap<String, String> allRequestParams,
+      @Parameter(
+              description =
+                  "Optional JSON request body containing search parameters. Used to bypass URL length limitations.")
+      @RequestBody(required = false)
+          Map<String, String> searchParamsBody);
+
+  // Backward-compatible overload when no request body is provided.
+  default List<Channel> query(MultiValueMap<String, String> allRequestParams) {
+    return query(allRequestParams, null);
+  }
 
   @Operation(
       summary = "Combined query for channels",
       description =
-          "Query for a collection of Channel instances and get a count and the first 10k hits.",
+          "Query for a collection of Channel instances and get a count and the first 10k hits. "
+              + "Search parameters can be provided via URL query string or JSON request body, or both. "
+              + "URL parameters take precedence for control parameters (~size, ~from, ~search_after, ~track_total_hits). "
+              + "Regular search parameters from URL and body are combined as separate values in the query.",
       operationId = "combinedQueryChannels",
       tags = {"Channel"})
   @ApiResponses(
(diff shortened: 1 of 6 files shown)


shroffk:
```
curl --location --request GET 'http://localhost:9090/ChannelFinder/resources/channels?iocid=127.0.0.1%3A*' \
--header 'Content-Type: application/json' \
--header 'Authorization: ••••••' \
--data '{"~name":"*74*,*75*", "hostName":"LEC-175107"}
```

now you are no longer limited by 600 chars

src/main/java/org/phoebus/channelfinder/web/v0/api/IChannel.java
```diff
@@ -77,11 +94,25 @@ List<Channel> query(
   @GetMapping("/combined")
   SearchResult combinedQuery(
       @Parameter(description = SEARCH_PARAM_DESCRIPTION) @RequestParam
-          MultiValueMap<String, String> allRequestParams);
+          MultiValueMap<String, String> allRequestParams,
```
jacomago, line 97: I think you can just do two methods rather than mutually exclusive inputs.

src/main/java/org/phoebus/channelfinder/web/v0/api/IChannel.java
```diff
@@ -77,11 +94,25 @@ List<Channel> query(
...
+      @Parameter(
+              description =
+                  "Optional JSON request body containing search parameters. Used to bypass URL length limitations.")
+      @RequestBody(required = false)
+          Map<String, String> searchParamsBody);
```
jacomago, line 102: I don't like a generic Map<String, String> better to use an actual record

record QueryOptions(String name, Map<String, String> properties, List<String> tags) {}

You can see it makes the swagger much much nicer.

jacomago (commented):
Do we really want to do this? Or make api version 2? Then we can get rid of the MultiValueMap.

src/main/java/org/phoebus/channelfinder/web/v0/api/IChannel.java
shroffk: well this is not part of the current change so I would address this as a separate issue

shroffk:
> Why should we do this now?

These small changes have significant impact on the performance of recceeiver.
Considering this is backward compatible I feel we should add this.

anderslindho:
I side with @jacomago, questioning if we should do this. Even if backwards compatible it feels the wrong direction. I think we instead should be adding these capabilities to the v1 API (but then as POST under the appropriate resource). 

This PR is also introducing some layer slippage: `SearchParameterMergerUtil` should prob go into `web/v0/` rather than `common/`. And I am not sure about the overload as test accommodation, plus the behaviour for query param PLUS body feels a bit strange. Finally, should we really be designing around current recsync stack when we have in mind to replace it?

shroffk:
> as POST under the appropriate resource

We can add to POST under v1 along with GET ( this is mirroring the behaviour of the elastic REST API )

>layer slippage: SearchParameterMergerUtil should prob go into web/v0/ rather than common/.

I thought since this is utility class, common would be the right place for it. We might have to expand it based on new requirements from v1. I am fine moving it to common.

> query param PLUS body feels a bit strange.

While I don't totally disagree, in v1 I am fine with splitting these into different end points

GET/POST ...ChannelFinder/resources/channels/_search/{channelName}

GET/POST ...ChannelFinder/resources/channels/_search + body with search criteria

> Finally, should we really be designing around current recsync stack when we have in mind to replace it?

These are the feature I need to complete my new implementation for recceiver. I need some of these changes to support the case of recceiver supporting millions of PV's

shroffk:
Considering the changes are backward compatible could we merge this and the batch delete... I don't think I will need any other changes and we can start focusing on v1

anderslindho:
I am not completely following @shroffk. I suggested POST because GET has a character limitation that makes it unfit for batch/bulk ops. I don't love the idea to mirror the ES REST API here; that would be exposing repo layer details in our API design, which is the opposite direction we want to go IMO. In fact I am not sure about the ES repo layer in CF in the first place, but that's a separate conversation.

Re recceiver2: we are 2 sites doing separate implementations and we have not in any way aligned on an improved design of either. I understand you need this for your new recceiver, but I don't think that should drive CF mainline API changes - especially not ones with known design issues. CF has its own lifecycle and multiple clients beyond recsync.

What I want us to move towards is a CF with proper clean layer separation and a truly RESTful API. Recsync is then a separate matter; it's just one way to populate CF, not the other way around.

For that reason I'd rather we hold this and do it properly in v1 as a POST endpoint under the right resource. I'm happy to prioritise that discussion if it unblocks you.

shroffk:
> that would be exposing repo layer details in our API design

I just used the elastic API as a reference/resources...  maybe mirroring was not a good word. This does not bind us to es or in anyway expose the underline storage layer.

I am in agreement with POST with a query body being the preferred solution for CF v1.

However, for the time being I just need these two updates to significantly simplify my workflow. The "new" recceiver was started a few years ago so I thought it would be nice to complete it.

I agree that CF has its own life cycle, but the recceiever is by far the most important client. It is the one which will usually defines the performance and scalability boundaries of CF.

I am guessing we will not be merging this so I will update my client... are we Ok with the batch delete? #213 #214

anderslindho:
> I am guessing we will not be merging this so I will update my client... are we Ok with the batch delete? https://github.com/ChannelFinder/ChannelFinderService/issues/213 https://github.com/ChannelFinder/ChannelFinderService/pull/214

Will comment in that PR for traceability

Issue #213, opened 2026-05-04
Add support to delete multiple channels in a single API call


shroffk:
Currently, deleting multiple channels requires individual calls to the single-channel delete endpoint. This results in highly inefficient workflows when dealing with large numbers of channels (hundreds of thousands in some cases).

## Background

The ability to delete multiple channels as a single API call was intentionally not supported in the past, citing safety concerns. However, after reviewing various real-world use cases — including:

- Recsync/Receiver workflows
- Unit and integration tests
- Phoebus client actions

— it is clear that bulk channel deletion is a commonly needed and currently missing operation.

## Problem

Without bulk delete support, clients must loop over individual channel delete calls. This is:

- **Inefficient**: N HTTP round-trips for N channels
- **Slow**: Especially painful when removing hundreds of thousands of channels
- **Error-prone**: Partial failures are harder to handle across many individual requests

## Proposed Solution

Add a new endpoint (or extend an existing one) to support deleting multiple channels in a single API call. To address the original safety concerns, access to this endpoint should be **restricted to users with a specific role** (e.g., `CF-ADMIN`), ensuring only authorized users can perform bulk deletions.

shroffk:
While initially it was suggested that we use `CF-ADMIN`
This is still a channel operation, so I would like to start with `CF_CHANNEL`

PR #214, opened 2026-05-04
Batch delete

shroffk:
Addresses issue #213 

This PR should be merged after #212

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/common/SearchParameterMergerUtil.java
+++ b/src/main/java/org/phoebus/channelfinder/common/SearchParameterMergerUtil.java
@@ -1,0 +1,90 @@
+package org.phoebus.channelfinder.common;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.util.MultiValueMap;
+
+/**
+ * Utility class for merging search parameters from URL query string and JSON request body.
+ *
+ * <p>Merging strategy:
+ *
+ * <ul>
+ *   <li>For regular search parameters (e.g., ~name, property names, ~tag): values from both URL and
+ *       body are added as separate entries in the MultiValueMap.
+ *   <li>For control parameters (~size, ~from, ~search_after, ~track_total_hits): URL values take
+ *       precedence over body values (body values ignored if URL value exists).
+ * </ul>
+ */
+public final class SearchParameterMergerUtil {
+
+  private static final String CONTROL_SIZE = "~size";
+  private static final String CONTROL_FROM = "~from";
+  private static final String CONTROL_SEARCH_AFTER = "~search_after";
+  private static final String CONTROL_TRACK_TOTAL_HITS = "~track_total_hits";
+
+  private SearchParameterMergerUtil() {
+    // Utility class - private constructor
+  }
+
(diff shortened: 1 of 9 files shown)


anderslindho:
This is based on #212 which we still are discussing, so I will ignore those commits.

- The HTTP codes seem wrong: we cannot have 404 here (DELETE /channels exists), and I am hesitant also over 500 on ES failure.
- I am not sure about returning bare (long) count from bulk DELETE, but maybe fine ?
- I don't entirely follow why 29389760fcd0b0e8232f64a16a5ca77b8c75b4de used `Refresh.True` for `deleteAllById`.
- Should access on bulk delete be ROLES.CF_CHANNEL? The linked issue says "To address the original safety concerns, access to this endpoint should be restricted to users with a specific role (e.g., CF-ADMIN), ensuring only authorized users can perform bulk deletions."
- e68dbe4402b56a82c0bb4b4260666ea51da139f4 is nice but should maybe have been separate PR; seems to deserve it.

From discussion in #212 
> However, for the time being I just need these two updates to significantly simplify my workflow. The "new" recceiver was started a few years ago so I thought it would be nice to complete it.

I am still quite hesitant over us including code here in main CF for local development needs. I am fine with it if the extension/code change(s) themselves make sense, but surely there are better ways to improve local dev workflow?

shroffk:
I can remove 404, maybe a 400 bad request might be more suited for situations where we are trying to delete non-existing channels
I am not sure about the issue with 500? 

It is saveall that was using refresh.True.
`refresh.True` is an expensive operation, in this case I don't think we need immediate visibility of the changes so using `refresh.wait_for` might be enough. We can change it back to true or make it configurable.

I have updated the Issue, considering this is still a Channel operation I think the current role mapping is correct. It is also consistent with the other batch channel operations.

> main CF for local development needs.

These are not changes for local development needs.
Should I bother cleaning up and opening new PR's or is this one a hard no also?

anderslindho:
> I can remove 404, maybe a 400 bad request might be more suited for situations where we are trying to delete non-existing channels

I think 400 also is wrong as it signifies malformed syntax, which this is not. Unfortunately the RFCs don't cover batch, or rather synchronous batch i.e. bulk, but the common three strategies here are:
* idempotency and silent success 20x - doesn't matter if it didn't exist, it now also does not exist; microsoft and others does this
* multi-status 207 (maybe the right fit here if we treat this as non-atomic; we can technically fail for different reasons on different names)
* atomic fail (basically how you treat it now) 422

> I am not sure about the issue with 500?

My thought process was that it maybe ought to have been a 502 or 503 instead. I am not sure. I think maybe we really should have _both_ 500 and 503 with Retry-After - the latter for temporary ES failure and fallback 500.

> Should I bother cleaning up and opening new PR's or is this one a hard no also?

I am happy with adding the bulk delete feature itself, but I think the point about package location from the linked PR applies also here, and I do not like the inclusion of the additive mix of query params and request body handling. I would personally prefer to see https://github.com/ChannelFinder/ChannelFinderService/commit/e68dbe4402b56a82c0bb4b4260666ea51da139f4 split out but it's not a blocker.

anderslindho:
But maybe @jacomago wants to chime in before more work is invested? (He is in training so might be preoccupied.)

shroffk:
I will split these out into individual PR's

PR #216, opened 2026-05-07
refactor: enforce clean layer separation across repository, service, and controller    

anderslindho:
Pushes the codebase toward cleaner layering: repositories are pure persistence adapters, services have no web framework dependencies, and the HTTP boundary is explicit.                                                                                                                                         
                                                            
Also adds a v0 DTO layer ahead of a planned v1 API, so the two can evolve independently without touching the service or domain layers.

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
+++ b/src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
@@ -65,6 +65,8 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.data.repository.CrudRepository;
 import org.springframework.stereotype.Repository;
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.util.MultiValueMap;
 
 // Jackson 2 required by elasticsearch-java 8.x JacksonJsonpMapper — migrate with ES 9
 
@@ -458,7 +460,7 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
 
   @Override
   public long count() {
-    return this.count(Map.of());
+    return this.count(new LinkedMultiValueMap<>());
   }
 
   /**
@@ -527,7 +529,7 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
    * @param searchParameters channel search parameters
    * @return matching channels
    */
-  public SearchResult search(Map<String, List<String>> searchParameters) {
+  public SearchResult search(MultiValueMap<String, String> searchParameters) {
     BuiltQuery builtQuery = getBuiltQuery(searchParameters);
     Integer finalSize = builtQuery.size;
     Integer finalFrom = builtQuery.from;
@@ -569,7 +571,7 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
     }
   }
 
-  private BuiltQuery getBuiltQuery(Map<String, List<String>> searchParameters) {
+  private BuiltQuery getBuiltQuery(MultiValueMap<String, String> searchParameters) {
     BoolQuery.Builder boolQuery = new BoolQuery.Builder();
     int size = esService.getES_QUERY_SIZE();
     int from = 0;
@@ -737,7 +739,7 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
    * @param searchParameters channel search parameters
    * @return count of the number of matches to the provided query
    */
-  public long count(Map<String, List<String>> searchParameters) {
+  public long count(MultiValueMap<String, String> searchParameters) {
     BuiltQuery builtQuery = getBuiltQuery(searchParameters);
 
     try {
@@ -765,7 +767,9 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
    * @return count of the number of matches to the provided query
    */
   public long countByProperty(String propertyName, String propertyValue) {
-    return this.count(Map.of(propertyName, List.of(propertyValue == null ? "*" : propertyValue)));
+    MultiValueMap<String, String> params = new LinkedMultiValueMap<>();
+    params.add(propertyName, propertyValue == null ? "*" : propertyValue);
+    return this.count(params);
   }
 
   /**
@@ -775,7 +779,9 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
    * @return count of the number of matches to the provided query
    */
   public long countByTag(String tagName) {
-    return this.count(Map.of("~tag", List.of(tagName)));
+    MultiValueMap<String, String> params = new LinkedMultiValueMap<>();
+    params.add("~tag", tagName);
+    return this.count(params);
   }
 
   /**
@@ -786,7 +792,7 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
    * @param searchParameters channel search parameters
    * @return next page with its cursor
    */
-  public Scroll scroll(String scrollId, Map<String, List<String>> searchParameters) {
+  public Scroll scroll(String scrollId, MultiValueMap<String, String> searchParameters) {
     BuiltQuery builtQuery = getBuiltQuery(searchParameters);
     try {
       SearchRequest.Builder builder =
--- a/src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
+++ b/src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
@@ -41,10 +41,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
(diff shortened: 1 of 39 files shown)


src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
```diff
@@ -512,7 +510,7 @@ public void deleteAll() {
    * @param searchParameters channel search parameters
    * @return matching channels
    */
-  public SearchResult search(MultiValueMap<String, String> searchParameters) {
+  public SearchResult search(Map<String, List<String>> searchParameters) {
```
georgweiss: Why replacing MultiValuedMap?

georgweiss (commented):
Added in-line comment

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
anderslindho: See commit message for that change:
> Services must not depend on Spring web binding types. Replace
> MultiValueMap<String, String> with Map<String, List<String>> in all
> service method signatures and repository public APIs. Controllers pass
> their MultiValueMap directly - no conversion needed since MultiValueMap
> IS-A Map<String, List<String>>.

georgweiss: ```MultiValuedMap``` is a class in spring framework core, not in a web library.

anderslindho: Good catch, misleading commit message. I still think it is an improvement as it is pure Java, which I think generally is preferably for repo and service layers. I will fix the commit message for now.

georgweiss: Disagree. Why should a service layer be "pure Java"? Spring offers a lot of APIs that make perfect sense in a service layer. Or: it would be quite limiting to restrict a service layer to only JDK APIs.

src/main/java/org/phoebus/channelfinder/exceptions/StorageException.java
```diff
@@ -0,0 +1,12 @@
+package org.phoebus.channelfinder.exceptions;
+
+public class StorageException extends RuntimeException {
```
jacomago: Feel like there should be a better name

RepositoryException?

src/main/java/org/phoebus/channelfinder/web/v0/mapper/ChannelMapper.java
```diff
@@ -0,0 +1,36 @@
+package org.phoebus.channelfinder.web.v0.mapper;
+
+import java.util.List;
+import java.util.stream.StreamSupport;
+import org.phoebus.channelfinder.entity.Channel;
+import org.phoebus.channelfinder.web.v0.dto.ChannelDto;
+
+public final class ChannelMapper {
```
jacomago, line 8: Why do we need a mapper class? Can we not do the mapping on the types themselves?

i.e.

Channel.toDto
ChannelDto.toChannel 

?

Or the otherway round

Channel.fromDto
ChannelDto.fromChannel

src/test/java/org/phoebus/channelfinder/ChannelControllerIT.java
```diff
@@ -562,4 +567,24 @@ void setupAll() {
   void tearDown() throws IOException {
     ElasticConfigIT.teardown(esService);
   }
+
```
jacomago: These methods feel kind of weird to be in test file.

src/main/java/org/phoebus/channelfinder/web/v0/mapper/ChannelMapper.java
anderslindho: Probably because my mind was on MapStruct and I did not think of that 🫣 

`Channel.toDto` sounds wrong - domain layer would depend on controller layer which is the wrong way around. So it'd have to be more like
```
ChannelDto.from
ChannelDto.toDomain
```
But it's not immediately clear to me if this will fit later on with v1 API. And it's to prep for that that I set up these (currently unnecessary) mappers in the first place.

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
anderslindho: Fair point. To be more clear, I the benefit is that the signature is not tied to a Spring interface - even if Spring is used elsewhere in the same (or lower) layer(s) for other reasons. And I guess the counter-argument is that `MultiValueMap` communicates intent more explicitly and is available either way.

TBH I have no strong feelings about this particular change, so I can drop it if you prefer it one way over the other.

src/main/java/org/phoebus/channelfinder/web/v0/mapper/ChannelMapper.java
jacomago: I think I would rather the conversion happens inside of the service layer, since that is the mapping from controller layer to repository.

src/main/java/org/phoebus/channelfinder/repository/PropertyRepository.java
```diff
@@ -142,7 +141,7 @@ public <S extends Property> S save(String propertyName, S property) {
     } catch (Exception e) {
       String message = MessageFormat.format(TextUtil.FAILED_TO_INDEX_PROPERTY, property.toLog());
       logger.log(Level.SEVERE, message, e);
-      throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, message, null);
+      throw new StorageException(message);
```
simon-ess: This feels a bit too broad compared to the others. Should we narrow this to only be ElasticExceptions and have a second catch block?

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
jacomago: I think maybe a better fix in the repo layer is to move the BuiltQuery construction up to the Service layer. It is really just a kind of wrapper object. 

For service layer I'm mixed, I don't like that its used in the controller layer but it is, so we can't get around it very much.

anderslindho: > I think maybe a better fix in the repo layer is to move the BuiltQuery construction up to the Service layer. It is really just a kind of wrapper object.

I don't think that makes architectural sense - BuiltQuery is a private record that assembles an Elasticsearch BoolQuery. Moving it up would either expose ES internals to the service or require making BuiltQuery public.

I will just revert the MultiValueMap conversion since that is contentious.

simon-ess:
Also, now that you have reverted one commit... maybe just remove both commits from the PR?

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
```diff
@@ -218,7 +217,7 @@ public Channel save(String channelName, Channel channel) {
     } catch (Exception e) {
       String message = MessageFormat.format(TextUtil.FAILED_TO_INDEX_CHANNEL, channel.toLog());
       logger.log(Level.SEVERE, message, e);
-      throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, message, null);
+      throw new RepositoryException(message, e);
```
simon-ess: Here's another bare `Exception`...

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
```diff
@@ -566,7 +561,7 @@ public SearchResult search(MultiValueMap<String, String> searchParameters) {
       String message =
           MessageFormat.format(TextUtil.SEARCH_FAILED_CAUSE, searchParameters, e.getMessage());
       logger.log(Level.SEVERE, message, e);
-      throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, message, e);
+      throw new RepositoryException(message, e);
     }
```
simon-ess: Another

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
```diff
@@ -754,7 +749,7 @@ public long count(MultiValueMap<String, String> searchParameters) {
       String message =
           MessageFormat.format(TextUtil.COUNT_FAILED_CAUSE, searchParameters, e.getMessage());
       logger.log(Level.SEVERE, message, e);
-      throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, message, e);
+      throw new RepositoryException(message, e);
     }
```
simon-ess: Another...

simon-ess (commented):
There are still several bare Exceptions being caught and rethrown. There may be more, I stopped looking after 3.

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
anderslindho: I think all are fixed now. There was some pre-existing debt here, so please check the new breakdown.

simon-ess:
```console
$ rg "catch\s*\(Exception\b" -g '!*test*'
src/main/java/org/phoebus/channelfinder/service/ChannelProcessorService.java
124:                      } catch (Exception e) {

src/main/java/org/phoebus/channelfinder/service/external/ArchiverService.java
115:    } catch (Exception e) {
138:    } catch (Exception e) {
163:    } catch (Exception e) {
264:    } catch (Exception e) {

src/main/java/org/phoebus/channelfinder/configuration/AAChannelProcessor.java
157:                      } catch (Exception e) {
```

Looks good enought to me.

simon-ess:
I'd approve it but I do not seem to have permission to do so.

jacomago (approved):

PR #217, opened 2026-05-07
Batch delete

shroffk:
Similar to PR #214 

I removed the 404 error

I made this end point "best effort", the response is the number of channels Cf successfully deleted.

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
+++ b/src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
@@ -293,7 +293,7 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
                 }
                 BulkResponse result;
                 try {
-                  result = client.bulk(br.refresh(Refresh.True).build());
+                  result = client.bulk(br.refresh(Refresh.WaitFor).build());
                   // Log errors, if any
                   if (result.errors()) {
                     logger.log(Level.SEVERE, TextUtil.BULK_HAD_ERRORS);
@@ -806,8 +806,39 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
 
   @Override
   public void deleteAllById(Iterable<? extends String> ids) {
-    // TODO Auto-generated method stub
+    List<String> idList =
+        StreamSupport.stream(ids.spliterator(), false)
+            .filter(id -> id != null && !id.isBlank())
+            .map(String::valueOf)
+            .distinct()
+            .toList();
 
+    for (int i = 0; i < idList.size(); i += chunkSize) {
+      List<String> chunk = idList.stream().skip(i).limit(chunkSize).toList();
+      BulkRequest.Builder br = new BulkRequest.Builder();
+      for (String id : chunk) {
+        br.operations(op -> op.delete(del -> del.index(esService.getES_CHANNEL_INDEX()).id(id)));
+      }
+      br.refresh(Refresh.True);
+
+      try {
+        BulkResponse result = client.bulk(br.build());
+        if (result.errors()) {
+          String message = MessageFormat.format(TextUtil.FAILED_TO_DELETE_CHANNEL, chunk);
+          logger.log(Level.SEVERE, TextUtil.BULK_HAD_ERRORS);
+          for (BulkResponseItem item : result.items()) {
+            if (item.error() != null) {
+              logger.log(Level.SEVERE, () -> item.error().reason());
+            }
+          }
+          throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, message, null);
+        }
+      } catch (IOException e) {
+        String message = MessageFormat.format(TextUtil.FAILED_TO_DELETE_CHANNEL, chunk);
+        logger.log(Level.SEVERE, message, e);
+        throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, message, e);
+      }
+    }
   }
 
   @PreDestroy
--- a/src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
+++ b/src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
@@ -433,7 +433,7 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
         return Collections.emptyList();
       }
 
-      int lookupBatchSize = Math.max(1, Math.min(chunkSize, esService.getES_QUERY_SIZE()));
+      int lookupBatchSize = Math.clamp(chunkSize, 1, esService.getES_QUERY_SIZE());
       List<Channel> result = new ArrayList<>();
 
       for (int i = 0; i < ids.size(); i += lookupBatchSize) {
--- a/src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
+++ b/src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
@@ -427,12 +427,7 @@ public class ChannelRepository implements CrudRepository<Channel, String> {
   @Override
   public List<Channel> findAllById(Iterable<String> channelIds) {
     try {
-      List<String> ids =
-          StreamSupport.stream(channelIds.spliterator(), false)
-              .filter(id -> id != null && !id.isBlank())
-              .collect(Collectors.toCollection(LinkedHashSet::new))
-              .stream()
-              .toList();
+      List<String> ids = normalizeIds(channelIds);
 
       if (ids.isEmpty()) {
         return Collections.emptyList();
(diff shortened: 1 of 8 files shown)


shroffk:
@jacomago I had to update one of your Integration tests.
I am not sure how the past tests were working... the expected exception was 500, but the deprecated V0ExceptionHandler maps it to 400

I think it makes sense for this to be 400 since this is a bad request (a null channel name)

tynanford (dismissed):

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
```diff
@@ -428,16 +428,33 @@ public Iterable<Channel> findAll() {
   public List<Channel> findAllById(Iterable<String> channelIds) {
     try {
       List<String> ids =
-          StreamSupport.stream(channelIds.spliterator(), false).collect(Collectors.toList());
+          StreamSupport.stream(channelIds.spliterator(), false)
+              .filter(id -> id != null && !id.isBlank())
```
jacomago: Could use StringUtils.isEmpty ?

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
```diff
@@ -806,8 +823,63 @@ public Scroll scroll(String scrollId, MultiValueMap<String, String> searchParame
 
   @Override
   public void deleteAllById(Iterable<? extends String> ids) {
-    // TODO Auto-generated method stub
+    List<String> idList =
+        StreamSupport.stream(ids.spliterator(), false)
```
jacomago: Pull out this code as a static filterOutEmptyIds as it is duplicated.

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
```diff
@@ -806,8 +823,63 @@ public Scroll scroll(String scrollId, MultiValueMap<String, String> searchParame
...
+          if (item.error() != null) {
+            logger.log(
+                Level.SEVERE,
+                () ->
+                    MessageFormat.format(
+                        "Failed to delete channel id {0}: {1}", item.id(), item.error().reason()));
+            continue;
+          }
+          if (Integer.valueOf(200).equals(item.status())) {
+            deletedCount++;
```
jacomago, line 854: Shouldn't this be deletedCount+chunk.size() ?

src/main/java/org/phoebus/channelfinder/service/ChannelService.java
```diff
@@ -179,6 +180,36 @@ public void remove(String channelName) {
...
 
+  public long remove(Iterable<String> channelNames) {
+    requireRole(ROLES.CF_CHANNEL, "channels batch");
+
+    List<String> distinctChannelNames =
+        StreamSupport.stream(channelNames.spliterator(), false)
+            .filter(name -> name != null && !name.isBlank())
+            .collect(Collectors.toCollection(LinkedHashSet::new))
+            .stream()
+            .toList();
```
jacomago: Again repeated code.

src/main/java/org/phoebus/channelfinder/service/ChannelService.java
```diff
@@ -179,6 +180,36 @@ public void remove(String channelName) {
...
+
+    if (distinctChannelNames.isEmpty()) {
+      return 0;
+    }
+
+    List<Channel> existingChannels = channelRepository.findAllById(distinctChannelNames);
```
jacomago: You don't need the filter on line 186, you added the filter in this method already.

src/main/java/org/phoebus/channelfinder/web/v0/api/IChannel.java
```diff
@@ -269,4 +269,24 @@ long queryCount(
...
+      value = {
+        @ApiResponse(responseCode = "200", description = "Number of channels deleted"),
+        @ApiResponse(
+            responseCode = "401",
+            description = "Unauthorized",
+            content = @Content(schema = @Schema(implementation = ResponseStatusException.class))),
+        @ApiResponse(
+            responseCode = "500",
+            description = "Error while trying to delete channels",
+            content = @Content(schema = @Schema(implementation = ResponseStatusException.class)))
```
jacomago, line 288: What about if a channel name doesn't exist? Should we respond with 404?

src/test/java/org/phoebus/channelfinder/service/ChannelServiceTest.java
```diff
@@ -138,4 +143,38 @@ void createChannel_validChannelWithTagAndProperty_noException() {
 
     assertDoesNotThrow(() -> channelService.create("ch", channel));
   }
+
```
jacomago, line 146: Could do with repository layer tests as well.

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
shroffk: Are you referencing the springboot StringUtils? cause the is empty is deprecated?
The apache commons one would need a new external dependency.

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
shroffk: ok, 
I will look into cleaning up the filtering... trying to decide where the responsiblity lies in the repository or the service

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
shroffk: ~~ah, good catch~~
thanks

shroffk: Actually,
this is correct
```
        BulkResponse result = client.bulk(br.build());
        for (BulkResponseItem item : result.items()) {
          if (item.error() != null) {
            logger.log(
                Level.SEVERE,
                () ->
                    MessageFormat.format(
                        "Failed to delete channel id {0}: {1}", item.id(), item.error().reason()));
            continue;
          }
          if (Integer.valueOf(200).equals(item.status())) {
            deletedCount++;
          }
        }
```
The result.items() will return one result for *each* of the delete requests which are packaged together in the bulk request 
`Required - The result of each operation in the bulk request, in the order they were submitted.`

The count should be right

src/main/java/org/phoebus/channelfinder/web/v0/api/IChannel.java
shroffk: I had considered that, but anders mentioned how 404 is not correct by rest standards since the end point doe exist.

For the time being this is best effort delete so the burden is shifted to the user to ensure things are correctly deleted even in the case of no errors from teh service.

src/main/java/org/phoebus/channelfinder/repository/ChannelRepository.java
jacomago: Ah I thought apache-commons was in. So nevermind.

src/main/java/org/phoebus/channelfinder/web/v0/api/IChannel.java
jacomago: Yeah I kind of agree.

src/test/java/org/phoebus/channelfinder/service/ChannelServiceTest.java
shroffk: Added tests for most of the combinations of input to the batch delete

shroffk: Added tests for the repository layer

src/test/java/org/phoebus/channelfinder/ChannelRepositoryIT.java
```diff
@@ -380,6 +380,127 @@ void deleteXmlTag() {
         "Failed to delete the channel");
   }
 
+  /** delete multiple channels */
+  @Test
+  void deleteXmlChannels() {
```
jacomago: Xml???

jacomago (commented):
Maybe fix some of the new sonar issues?

src/test/java/org/phoebus/channelfinder/ChannelRepositoryIT.java
shroffk: following the naming convention of the other tests

shroffk: there is deleteXmlTag, saveXmlChannel..... almost all the tests are named such

shroffk:
> Maybe fix some of the new sonar issues?

fixed 2, 
the third one is a todo which is beyond the scope of this PR, we need to have more agreement on that

jacomago (approved):

PR #220, opened 2026-05-18
improve the channel validation pipeline

shroffk:
- Refactored validation in `ChannelService` to reduce duplication:
  - `validateChannel(...)` now routes through `validateChannels(List.of(channel))`
  - all channel validation logic is centralized in one flow ( previous there were 2 pathways with a bunch of duplicate code )

- Improved validation efficiency:
  - `validateChannels(...)` loads existing tags/properties once per call
  - each channel is validated against these in-memory snapshots
  - **avoids repeated Elasticsearch existence checks (`existsById`) for every tag/property**

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/service/ChannelService.java
+++ b/src/main/java/org/phoebus/channelfinder/service/ChannelService.java
@@ -184,74 +184,63 @@ public class ChannelService {
         .collect(Collectors.toMap(Channel::getName, c -> c));
   }
 
+  /**
+   * Validates one channel payload.
+   *
+   * <p>Validation requires a non-empty channel name and owner, verifies that all referenced tags
+   * and properties exist, and rejects null/empty property values.
+   *
+   * @param channel channel to validate
+   */
   private void validateChannel(Channel channel) {
-    if (channel.getName() == null || channel.getName().isEmpty()) {
-      throw new ChannelValidationException(
-          MessageFormat.format(TextUtil.CHANNEL_NAME_CANNOT_BE_NULL_OR_EMPTY, channel.toLog()));
-    }
-    if (channel.getOwner() == null || channel.getOwner().isEmpty()) {
-      throw new ChannelValidationException(
-          MessageFormat.format(TextUtil.CHANNEL_OWNER_CANNOT_BE_NULL_OR_EMPTY, channel.toLog()));
-    }
-    for (Tag tag : channel.getTags()) {
-      if (!tagRepository.existsById(tag.getName())) {
-        throw new TagNotFoundException(tag.getName());
-      }
-    }
-    checkPropertyValues(channel);
+    validateChannels(List.of(channel));
   }
 
+  /**
+   * Validates each channel payload in the provided iterable.
+   *
+   * <p>For every channel, validation requires a non-empty name and owner, verifies that all
+   * referenced tags and properties exist, and rejects null/empty property values.
+   *
+   * @param channels channels to validate
+   */
   private void validateChannels(Iterable<Channel> channels) {
-    List<String> existingProperties =
-        StreamSupport.stream(propertyRepository.findAll().spliterator(), true)
+    Iterable<Tag> existingTags = tagRepository.findAll();
+    List<String> existingTagNames =
+        StreamSupport.stream(existingTags.spliterator(), true).map(Tag::getName).toList();
+
+    Iterable<Property> existingProperties = propertyRepository.findAll();
+    List<String> existingPropertyNames =
+        StreamSupport.stream(existingProperties.spliterator(), true)
             .map(Property::getName)
             .toList();
-    List<String> existingTags =
-        StreamSupport.stream(tagRepository.findAll().spliterator(), true)
-            .map(Tag::getName)
-            .toList();
+
     for (Channel channel : channels) {
-      validateChannelAgainst(channel, existingTags, existingProperties);
+      if (channel.getName() == null || channel.getName().isEmpty()) {
+        throw new ChannelValidationException(
+            MessageFormat.format(TextUtil.CHANNEL_NAME_CANNOT_BE_NULL_OR_EMPTY, channel.toLog()));
+      }
+      if (channel.getOwner() == null || channel.getOwner().isEmpty()) {
+        throw new ChannelValidationException(
+            MessageFormat.format(TextUtil.CHANNEL_OWNER_CANNOT_BE_NULL_OR_EMPTY, channel.toLog()));
+      }
+      for (Tag tag : channel.getTags()) {
+        if (!existingTagNames.contains(tag.getName())) {
+          throw new TagNotFoundException(tag.getName());
+        }
+      }
+      List<String> propNames = channel.getProperties().stream().map(Property::getName).toList();
+      for (String propName : propNames) {
+        if (!existingPropertyNames.contains(propName)) {
+          throw new PropertyNotFoundException(propName);
+        }
+      }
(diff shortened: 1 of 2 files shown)


src/main/java/org/phoebus/channelfinder/service/ChannelService.java
```diff
@@ -184,74 +184,63 @@ private Map<String, Channel> findExistingChannels(List<Channel> channels) {
...
+   * <p>For every channel, validation requires a non-empty name and owner, verifies that all
+   * referenced tags and properties exist, and rejects null/empty property values.
+   *
+   * @param channels channels to validate
+   */
   private void validateChannels(Iterable<Channel> channels) {
-    List<String> existingProperties =
-        StreamSupport.stream(propertyRepository.findAll().spliterator(), true)
+    Iterable<Tag> existingTags = tagRepository.findAll();
+    List<String> existingTagNames =
```
jacomago: I think better to make this a hashset.

src/main/java/org/phoebus/channelfinder/service/ChannelService.java
```diff
@@ -184,74 +184,63 @@ private Map<String, Channel> findExistingChannels(List<Channel> channels) {
...
+        StreamSupport.stream(existingTags.spliterator(), true).map(Tag::getName).toList();
+
+    Iterable<Property> existingProperties = propertyRepository.findAll();
+    List<String> existingPropertyNames =
```
jacomago: Also make this a hashset.

src/main/java/org/phoebus/channelfinder/service/ChannelService.java
```diff
@@ -184,74 +184,63 @@ private Map<String, Channel> findExistingChannels(List<Channel> channels) {
...
 
+  /**
+   * Validates each channel payload in the provided iterable.
+   *
+   * <p>For every channel, validation requires a non-empty name and owner, verifies that all
+   * referenced tags and properties exist, and rejects null/empty property values.
+   *
+   * @param channels channels to validate
+   */
   private void validateChannels(Iterable<Channel> channels) {
```
jacomago, line 209: As the sonar message suggests the complexity is still high in this method. I think we should split this up into

- validateNoEmptyChannels
- validatePropertiesExist
- validateTagsExist

And this just calls these other methods.

shroffk: I disagree... I am finding this explosion of code of splitting a simple function into multiple simpler functions does little to improve readability. Now you have to navigate a 4 functions to just understand how to validate a channel

shroffk: there is no other place where the functions would be used other than this one method, so why split it up now... if in the future we see that we need to only validate a tag or a property... then fine we can split these out

src/main/java/org/phoebus/channelfinder/service/ChannelService.java
shroffk: yes, 
this makes perfect sense.. making the changes now

shroffk:
@jacomago so other than the one sonarcube issue I don't see any others.
Is this ready to merge?

anderslindho:
If we choose to oppose sonar, can we at least approve/confirm in sonar or else add a NOSONAR comment with explanation for why we disagree? I don't quite like us adding to the issue count.

shroffk:
Just to clarify, 
I am only disagreeing in this **specific** case... This is not a blanket disagreement with all complexity issues raised by sonar

src/main/java/org/phoebus/channelfinder/service/ChannelService.java
jacomago: I disagree... I think it is difficult to read now. I just read it five times to realise how it fully works.

I think something like this would be good:

```java

for (channel: channels) {

validateNotEmpty(channel)
validateExistingTags(channe, existingTags)
validateExistingProperties(channel, existingProperties)
validatePropertyValues(channel)

}

```

But there is another problem, what if we want to check every channel and respond with the problems with all of them. Then maybe we should go with:

```java

class MutliChannelValidationException {

   MutliChannelValidationException(Map<String, Exception> channelErrors) {
   super(constructSomeNiceErrorMessage)

}

}

void validateChannels(channels) {
Map<String, Exception> errors = channels.stream().map(channel - >
Entry.of(channel.name, () ->
try {
validateNotEmpty(channel)
validateExistingTags(channe, existingTags)
validateExistingProperties(channel, existingProperties)
validatePropertyValues(channel)
} catch (ChannelValidationException ex) {
  logger.debug(ex)
  return ex
}
return null
}
)).filter(e -> e.getValue() != null).collect(Collectors.toMap);

if (!errors.isEmpty()) {
throw MutliChannelValidationException(errors)
}

```

I'm guessing you are doing this for performance, so I'm not actually sure which would be quicker since the stream version presumbably we could just make a parallelStream, but the other one cancels early?

shroffk: readability: 
maybe a few comments?
```
    for (Channel channel : channels) {
      // Check 1: Validate that channel name is non-null and non-empty
      if (channel.getName() == null || channel.getName().isEmpty()) {
        throw new ChannelValidationException(
            MessageFormat.format(TextUtil.CHANNEL_NAME_CANNOT_BE_NULL_OR_EMPTY, channel.toLog()));
      }

      // Check 2: Validate that channel owner is non-null and non-empty
      if (channel.getOwner() == null || channel.getOwner().isEmpty()) {
        throw new ChannelValidationException(
            MessageFormat.format(TextUtil.CHANNEL_OWNER_CANNOT_BE_NULL_OR_EMPTY, channel.toLog()));
      }

      // Check 3: Validate that all referenced tags exist in the ChannelFinder
      for (Tag tag : channel.getTags()) {
        if (!existingTagNames.contains(tag.getName())) {
          throw new TagNotFoundException(tag.getName());
        }
      }

      // Check 4: Validate that all referenced properties exist and have non-empty names and values
      List<String> propNames = channel.getProperties().stream().map(Property::getName).toList();
      for (String propName : propNames) {
        if (!existingPropertyNames.contains(propName)) {
          throw new PropertyNotFoundException(propName);
        }
      }
      List<String> propValues = channel.getProperties().stream().map(Property::getValue).toList();
      checkPropertyValues(propNames, propValues);
    }
  }
 ```

About the checking of multiple channels.
This is not the current contract and we don't have this behaviour for any of the other validation operations. I generally like the idea of giving a complete validatation response instead of just throwing an exception at the first issue...but this can be separate issue and then we can have it applied to all the validations

anderslindho: I think Sky's suggested way of addressing the complexity lint reads far better than current or commented. But the multi-channel validation/granular response feels a bit premature and scope creep-y here.

shroffk: fine!

anderslindho: We still seem to be getting an issue warning from sonar - would be nice if we could eliminate it.

shroffk: split up the methods,
sonarcube is not checking out the correct/last commit...

anderslindho: OK good, not just me going insane then.

anderslindho (approved):

Issue #53, opened 2022-05-13
How to handle property values which are empty strings


shroffk:
Properties with empty value should be treated as a delete

shroffk:
@jacomago this would address the clear properties during update requirement we were talking about.
This was a feature of older ChannelFinder services and I think we should add it back

jacomago:
I would say it doesn't sound very restful. Can't we use the owner to say, these are my properties and and I push all of them?

This using empty string doesn't work in the case that recceiver restarts with less managed properties than it first started with as it wouldn't know what properties it lost.

shroffk:
Ok,
So I would like to split these into 2

> Can't we use the owner to say, these are my properties and and I push all of them?

Could you help me better understand this, like an example pseudo code set of calls

jacomago:
Something like


```java

Response channelUpdate(Channel channel, String owner) {
   Channel currentChannel = getCurrentChannel(channel.name);
   Channel newChannel = currentChannel;
   for (Property property: channel.getProperties()) {
     newChannel.setPropertyValue(property.name, property.value);
  }
  
  // Remove owned properties
  Set<String> currentOwnedPropertyNames = currentChannel.getOwnedProperties(owner);
  Set<String> propertiesToRemove = currentOwnedPrope.removeAll(channel.getProperties().stream().map(e -> e.getKey).collect());
  newChannel.removeProperties(propertiesToRemove)
  
  repository.updateChannel(newChannel)
}

```

shroffk:
Hmm... 
I think I understand better., we are basically creating groups of properties ( grouped by owner ), which are then managed in a sort of collective fashion.

This does solves the issue of handling properties the client might not know about ( case: recceiver starts with fewer managed properties )

However, in this case the client would be performing an operation where they might not clearly know what is happening on the service side. The one benefit of the first option is that while not being pure REST it is very clear on the client side what is going to be added and removed ( no guessing which other properties will be modified )... but like you pointed out this would not be able to handle the reduced property set.

shroffk:
anyone with any other ideas? @tynanford

shroffk:
So, I like the idea of 

```
DELETE /ChannelFinder/resources/properties/archive/channels
Content-Type: application/json

["channela", "channelb", "channelc"]
```

I understand that this does not address all the issues but
- it is simple and clean
- fits with the existing service API ( yes I know it is using a body... but honestly I think it is fine )

jacomago:
What about

[...]

PR #222, opened 2026-05-21
Batch delete a property from multiple channels

shroffk:
#53


jacomago (commented):
Why not create api v2? or v1? 

I feel like trying to maintain the existing one like this is problematic. A couple of very small fixes in the next api:

- properties are requested and returned as a map
- Removal of the the channels from the property and tag response object

They would have so much impact on the quality of the client code.

I think moving on to creation of v1 now is worthwhile. We make another release and then start the next api rather than trying to massage this one into having all the features we want.

shroffk:
@copilot resolve the merge conflicts in this pull request

PR #223, opened 2026-05-22
AA Processor refactor

jacomago:
Refactor for the Archiver Channel Processor

- Simplify the code by making it more DRY
- Extract some lambdas for clarity
- Reduce INFO and WARNING level logging (by swapping to counts instead of full lists) and use lambda logging

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/configuration/AAChannelProcessor.java
+++ b/src/main/java/org/phoebus/channelfinder/configuration/AAChannelProcessor.java
@@ -107,19 +107,24 @@ public class AAChannelProcessor implements ChannelProcessor {
       return 0;
     }
 
-    // Get Info (policy and version) of each archiver
     Map<String, ArchiverInfo> archiversInfo = getArchiversInfo(aaURLs);
     if (archiversInfo.isEmpty()) {
       return 0;
     }
 
-    Map<String, List<ArchivePVOptions>> archiverAliasToArchivePVOptions =
-        new HashMap<>(); // AA identifier, ArchivePVOptions
-    for (String alias : archiversInfo.keySet()) {
-      archiverAliasToArchivePVOptions.put(alias, new ArrayList<>());
-    }
-
     logger.log(Level.INFO, "Get channelfinder properties for aa processor.");
+    Map<String, List<ArchivePVOptions>> pvsByArchiver = buildArchivePVMap(channels, archiversInfo);
+
+    long count = submitToArchivers(pvsByArchiver, archiversInfo);
+    logger.log(Level.INFO, () -> String.format("Configured %s channels.", count));
+    return count;
+  }
+
+  private Map<String, List<ArchivePVOptions>> buildArchivePVMap(
+      List<Channel> channels, Map<String, ArchiverInfo> archiversInfo) {
+    Map<String, List<ArchivePVOptions>> result = new HashMap<>();
+    archiversInfo.keySet().forEach(alias -> result.put(alias, new ArrayList<>()));
+
     channels.forEach(
         channel -> {
           Optional<Property> archiveProperty = findProperty(channel, archivePropertyName);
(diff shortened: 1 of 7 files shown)


src/main/java/org/phoebus/channelfinder/configuration/AAChannelProcessor.java
```diff
@@ -121,11 +121,7 @@ public long process(List<Channel> channels) throws JacksonException {
     logger.log(Level.INFO, "Get channelfinder properties for aa processor.");
     channels.forEach(
         channel -> {
-          Optional<Property> archiveProperty =
```
anderslindho, line 124: Hmm.. Commit message is wrong for ad9625291817264327d4296c79086b0508bd680f: it says you are eliminating **three** duplicate inline stream-filter-findFirst patterns, but you are just extracting two separate calls to two separate helpers. Don't really mind the refactor but there is nothing DRY about this (and the commit message is wrong).

src/main/java/org/phoebus/channelfinder/configuration/AAChannelProcessor.java
```diff
@@ -129,44 +134,41 @@ public long process(List<Channel> channels) throws JacksonException {
                     archiverAlias -> {
                       try {
                         addChannelChange(
-                            channel,
-                            archiverAliasToArchivePVOptions,
-                            archiversInfo,
-                            archiveProperty,
-                            archiverAlias);
+                            channel, result, archiversInfo, archiveProperty, archiverAlias);
```
anderslindho, line 145: Drop reformats like this from functional commits (45d60a8129e96d3c63e628aded70c37925fac270) - just annoying for reviewers

src/main/java/org/phoebus/channelfinder/configuration/AAChannelProcessor.java
```diff
@@ -14,6 +14,7 @@
 import org.apache.commons.lang3.StringUtils;
```
anderslindho, line 14: Isn't this commit, db2c390392a699559e9eb730051d64214b146dc6, really at the heart of the PR? IMO you should make this clear if that is the case. And I'd argue it should maybe even be factored out of this (otherwise refactor and log tuning) MR, and should describe the issues we've seen in prod at ESS today.

src/main/java/org/phoebus/channelfinder/service/external/ArchiverService.java
```diff
@@ -81,49 +81,34 @@ private Stream<List<String>> partitionSet(Set<String> pvSet, int pageSize) {
   public List<Map<String, String>> getStatuses(
       Map<String, ArchivePVOptions> archivePVS, String archiverURL, String archiverAlias) {
     Set<String> pvs = archivePVS.keySet();
-    boolean postSupportOverride = postSupportArchivers.contains(archiverAlias);
-    logger.log(Level.INFO, "Archiver Alias: {0}", archiverAlias);
-    logger.log(Level.INFO, "Post Support Override Archivers: {0}", postSupportArchivers);
```
georgweiss, line 86: Is this logging expendable?

georgweiss (commented):
One tiny comment in-lined

src/main/java/org/phoebus/channelfinder/service/external/ArchiverService.java
jacomago: This is now available by an endpoint processors/{processorID}/info via https://github.com/ChannelFinder/ChannelFinderService/pull/223/changes#diff-73d27f7214dad270850b445666d7c7fe83cb1232cea95c40420fe88b740f5936L90

georgweiss (dismissed):

anderslindho (commented):
Can we split up this MR? The mix of refactors, (pretty critical) bugfixes, and features is a little bit annoying. And we end up with some confusing categories - b2a99c17175ee037f8d26aae7356e9d6a6664bea is not a `fix` in the same way that the other `fix` commits are.

anderslindho (dismissed):
Lean mean merging machine 

Nitpick but calibrate your commit subject lengths - they are overflowing

jacomago:
> Can we split up this MR? The mix of refactors, (pretty critical) bugfixes, and features is a little bit annoying. And we end up with some confusing categories - [b2a99c1](https://github.com/ChannelFinder/ChannelFinderService/commit/b2a99c17175ee037f8d26aae7356e9d6a6664bea) is not a `fix` in the same way that the other `fix` commits are.

Split up, will do:

- refactor
- fixes
- policy feature
- tunable executor feature

anderslindho (approved):

PR #224, opened 2026-05-27
Give channel processors a tunable, named thread pool

jacomago:
Move the processor thread pool to its own component. 

Make it tunable by configuration.

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/Application.java
+++ b/src/main/java/org/phoebus/channelfinder/Application.java
@@ -31,9 +31,7 @@ import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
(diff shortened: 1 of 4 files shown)


shroffk (approved):

simon-ess (approved):
L(broadly)GTM, but is this a behaviour change to drop the oldest element of the queue instead of rejecting the newest one being added?

jacomago:
> L(broadly)GTM, but is this a behaviour change to drop the oldest element of the queue instead of rejecting the newest one being added?

Yes, but mostly I think channelfinder should be able to handle a high load on the processor and this was a bottleneck. This is where having the nin archiver have long processing (due to the 2 minutes timeout) was then causing a drop of the aa processing jobs. So I think my 'fixes' for the resume not working are multi strategy:

- Allow configuring this to have a larger pool (so it can handle having more requests at the same time) this pull request
- Add a short connection timeout (so down archivers have a smaller impact) https://github.com/ChannelFinder/ChannelFinderService/pull/225 
- Move the policy fetch to be a scheduled cache fetch (so fetching policies doesn't impact time to process channels.) Another pull request coming

PR #225, opened 2026-05-27
Aa Processor fixes

jacomago:
- Add a connection timeout 
- Skip archivers that are not available
- log a warning when pvs are not resumed

### DIFF
--- a/src/main/java/org/phoebus/channelfinder/configuration/AAChannelProcessor.java
+++ b/src/main/java/org/phoebus/channelfinder/configuration/AAChannelProcessor.java
@@ -14,6 +14,7 @@ import java.util.stream.Stream;
 import org.apache.commons.lang3.StringUtils;
 import org.phoebus.channelfinder.entity.Channel;
 import org.phoebus.channelfinder.entity.Property;
+import org.phoebus.channelfinder.exceptions.ArchiverServiceException;
 import org.phoebus.channelfinder.service.external.ArchiverService;
 import org.phoebus.channelfinder.service.model.archiver.ChannelProcessorInfo;
 import org.phoebus.channelfinder.service.model.archiver.aa.ArchiveAction;
@@ -182,8 +183,10 @@ public class AAChannelProcessor implements ChannelProcessor {
       }
       Map<String, ArchivePVOptions> pvMap =
           e.getValue().stream().collect(Collectors.toMap(ArchivePVOptions::getPv, pv -> pv));
-      count +=
-          archiverService.configureAA(getArchiveActions(pvMap, archiverInfo), archiverInfo.url());
+      Optional<Map<ArchiveAction, List<ArchivePVOptions>>> actions =
+          getArchiveActions(pvMap, archiverInfo);
+      if (actions.isEmpty()) continue;
+      count += archiverService.configureAA(actions.get(), archiverInfo.url());
     }
     return count;
   }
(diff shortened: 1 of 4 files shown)


anderslindho:
Can you please fix the 3 new sonar issues?

georgweiss (dismissed):

jacomago:
> Can you please fix the 3 new sonar issues?

done

anderslindho (commented):
Only thing I didn't like was the log of failed fetch N channels. But claude found more issues that I want you to look into:

  1. Null guard returns wrong Optional (AAChannelProcessor:271) — The archiverInfo == null check inside getArchiveActions returns Optional.of(Map.of()) (non-empty) instead of Optional.empty(). Currently
  dead code due to the outer guard in submitToArchivers, but if that guard is removed, the caller proceeds to archiverInfo.url() and NPEs silently instead of triggering the skip/warn path.
  2. Connect-timeout has no test (ArchiverServiceTest:374) — The new test accepts the TCP connection before hanging, so it exercises setReadTimeout, not setConnectTimeout. Deleting the new
  setConnectTimeout(...) line from production would leave all tests green.
  3. Double WARNING per failure (AAChannelProcessor:302) — A single ArchiverServiceException produces two separate Level.WARNING log entries (the getArchiveActions catch + warnSkippedActivePvs). Alerting
  systems counting by severity will double-fire.
  4. Misleading "RESUME was not sent" for non-paused PVs (AAChannelProcessor:292) — The warning includes all Active PVs, not just the ones that were paused; first-time-archive PVs get a misleading
  "remain paused" message.
  5. Deprecated setConnectTimeout(int) (ArchiverService:67) — PR adds a second call using the deprecated int form instead of fixing both to Duration.ofSeconds(timeoutSeconds).

jacomago:
> Only thing I didn't like was the log of failed fetch N channels. But claude found more issues that I want you to look into:
> 
>     1. Null guard returns wrong Optional (AAChannelProcessor:271) — The archiverInfo == null check inside getArchiveActions returns Optional.of(Map.of()) (non-empty) instead of Optional.empty(). Currently
>        dead code due to the outer guard in submitToArchivers, but if that guard is removed, the caller proceeds to archiverInfo.url() and NPEs silently instead of triggering the skip/warn path.
> 
Fixed
>     2. Connect-timeout has no test (ArchiverServiceTest:374) — The new test accepts the TCP connection before hanging, so it exercises setReadTimeout, not setConnectTimeout. Deleting the new
>        setConnectTimeout(...) line from production would leave all tests green.
> 
Added a test and made connection timeout separate, but not sure about the test since it uses a test IP which looks flaky. Will see if passes in CI.
>     3. Double WARNING per failure (AAChannelProcessor:302) — A single ArchiverServiceException produces two separate Level.WARNING log entries (the getArchiveActions catch + warnSkippedActivePvs). Alerting
>        systems counting by severity will double-fire.
> 
Dropped that commit.

>     4. Misleading "RESUME was not sent" for non-paused PVs (AAChannelProcessor:292) — The warning includes all Active PVs, not just the ones that were paused; first-time-archive PVs get a misleading
>        "remain paused" message.
> 
In commit that's dropped.
>     5. Deprecated setConnectTimeout(int) (ArchiverService:67) — PR adds a second call using the deprecated int form instead of fixing both to Duration.ofSeconds(timeoutSeconds).

I don't know why it thinks its deprecated: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/client/SimpleClientHttpRequestFactory.html

anderslindho:
Please fix introduced sonar issue 🙃

anderslindho (approved):

PR #226, opened 2026-05-29
Aa policy cache

jacomago:
This changes how the policies are fetched from the archiver. Instead on every "processing", it is scheduled to a default of every hour and the result cached. 

This separates the processing vs the fetching of policies from the archivers.


georgweiss (dismissed):

src/main/java/org/phoebus/channelfinder/service/model/archiver/aa/ArchiverPolicies.java
```diff
@@ -0,0 +1,10 @@
+package org.phoebus.channelfinder.service.model.archiver.aa;
+
+import java.util.List;
+
+public record ArchiverPolicies(List<String> policies) {
+  @Override
+  public String toString() {
+    return policies.size() + " policies";
```
simon-ess: Is this really the right `toString` implementation, just a count?

simon-ess (dismissed):

anderslindho:
Nitpick but please configure your agents on commit message subject lengths - you are again overflowing

src/test/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessorMultiArchiverIT.java
```diff
@@ -108,18 +108,6 @@ void testProcessMultiArchivers(
             .toList();
     when(archiverService.getStatusesViaGet(anyString(), anyList())).thenReturn(archivePVStatuses);
```
anderslindho, line 97: Nit: in commit (31c03d4ca95b11167aaecbc8810d01e2eb0b5b98) message "were left in the test from a copy-paste artifact" is irrelevant.

src/test/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessorMultiArchiverIT.java
```diff
@@ -132,7 +120,8 @@ void testProcessMultiArchivers(
     verify(archiverService, times(2)).getAAPolicies(anyString());
 
     if (!namesToStatuses.isEmpty()) {
-      verify(archiverService, times(1)).getStatusesViaGet(anyString(), anyList());
+      verify(archiverService).getStatusesViaGet(anyString(), anyList());
+      verify(archiverService).getStatusesViaPost(anyString(), anyList());
```
anderslindho, line 108: This change should be split out. It is a fix in what should be a no-op refactor commit. And the conventional commit category here (for the refactor) should prob be `refactor(tests)` rather than `test(...)`.

src/main/java/org/phoebus/channelfinder/configuration/AAChannelProcessor.java
```diff
@@ -92,7 +120,15 @@ public ChannelProcessorInfo processorInfo() {
             "Archivers",
             aaURLs.keySet().toString(),
             "AutoPauseOn",
-            autoPauseOptions.toString()));
+            autoPauseOptions.toString(),
+            "PostSupportArchivers",
+            postSupportArchivers.toString(),
+            "PolicyRefreshIntervalSeconds",
+            String.valueOf(policyRefreshIntervalSeconds),
```
anderslindho: These names should be improved, esp the first one (latter just add `Cache` somewhere)

src/test/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessorMultiArchiverIT.java
```diff
@@ -116,9 +120,7 @@ void testProcessMultiArchivers(
 
     aaChannelProcessor.process(channels);
 
-    // Verifications
-    verify(archiverService, times(2)).getAAPolicies(anyString());
-
+    // Verifications: query archiver uses GET, post archiver uses POST (aa.post_support=post)
```
anderslindho: That is how those methods work, yes...

src/main/java/org/phoebus/channelfinder/service/model/archiver/aa/ArchiverPolicies.java
jacomago: Hmm, yeah maybe it should be a toInfoString and have toString be the normal record implementation.

src/main/java/org/phoebus/channelfinder/configuration/AAChannelProcessor.java
jacomago: Tried to make more consistent.

src/test/java/org/phoebus/channelfinder/processors/aa/AAChannelProcessorMultiArchiverIT.java
jacomago: Not sure exactly what you mean.

anderslindho: Commit message for e2993540efe4e90f2f361cc04e8b722e6aeb300f is categorised as refactor. But the commit message body says `add missing post verification` (it was even more explicit before). Adding something missing is IMO not refactor but fix. The commit appears to contain 1 refactor 1 fix and should thus be split.

src/main/java/org/phoebus/channelfinder/configuration/AAChannelProcessor.java
anderslindho: I still think `PostSupportArchivers` could be improved

PR #227, opened 2026-06-02
Set the version to 5.1.0

jacomago:


anderslindho (approved):

shroffk:
Did you want to do a release? 
shouldn't you just release 5.1.0 and then set the version to the next snapshot?

Now with this change the maven release plugin will not manage this... we will have to always manually update the version before all releases

PR #231, opened 2026-07-06
docs(api): add API conventions, migrations, and README

jacomago:
Document API conventions including RSQL query syntax, bulk operations, jobs/batch definitions, and migration notes under dev-docs/api/.

### DIFF
--- a/dev-docs/api/conventions.md
+++ b/dev-docs/api/conventions.md
@@ -1,0 +1,531 @@
+# ChannelFinder REST API conventions (v1)
+
+Design rules for the v1 ChannelFinder HTTP API. They describe the target style,
+not the current API, which predates them.
+
+## Resources
+
+Domain resources:
+
+| Resource     | Path          | What                                                                          |
+| ------------ | ------------- | ---------------------------------------------------------------------------- |
+| channels     | `/channels`   | a unique `name`, the `sources` that claim it (active while any claim is), properties, tags |
+| properties   | `/properties` | a name-value definition (`name`), attached to channels with a per-channel `value` |
+| tags         | `/tags`       | a name-only label (`name`), attached to channels                             |
+
+Operational resources — addressable machinery, not directory data:
+
+| Resource   | Path                | What                                                                       |
+| ---------- | ------------------- | -------------------------------------------------------------------------- |
+| jobs       | `/jobs/{job_id}`    | a handle on a long-running async operation                                 |
+| processors | `/processors`       | channel post-processors; inspect/configure here, run with `POST /channels:process` |
+| sources    | `/sources`          | read-only source info of which producer claims what                            |
+
+Service metadata, health, and metrics are **not** resources — they live on the
+Spring Boot Actuator surface (`/actuator/*`), outside the versioned API.
+
+Channel serialization:
+
+- Properties serialize as one name→value map: `"properties": { "device": "bpm", "cell": 2, "position": 187.28 }`.
+- Tags serialize as a name list: `"tags": ["active"]`.
+- A property value is a **string**, an **integer**, or a **float** — fixed at the property's definition. JSON string vs. number separates text from numeric; the definition tells integer from float.
+- The per-channel datum is just the value; a property name cannot appear twice.
+
+## REST conventions
+
+CRUD maps to HTTP methods; anything that is not CRUD is a colon-verb command
+(see Commands).
+
+| Method   | Semantics                          | Success on write                                  | Idempotent |
+| -------- | ---------------------------------- | ------------------------------------------------- | ---------- |
+| `GET`    | reads, never mutates               | —                                                 | yes        |
+| `POST`   | creates                            | `201 Created` + full resource + `Location`        | no         |
+| `PUT`    | replaces with a full representation | `200 OK` + full resource                          | yes        |
+| `PATCH`  | updates part                       | `200 OK` + full resource                          | no         |
+| `DELETE` | removes                            | `204 No Content`                                  | yes        |
+
+No bare `201`/`200` — a caller never needs a follow-up `GET` for server-assigned
+fields.
+
+Status codes:
+
+| Code  | Meaning              | When                                                                                          |
+| ----- | -------------------- | --------------------------------------------------------------------------------------------- |
+| `200` | OK                   | `GET`; `PUT`/`PATCH` returning the full resource                                              |
+| `201` | Created              | `POST` persisting a resource; full resource + `Location`                                      |
+| `202` | Accepted             | async work accepted; `Location: /jobs/{job_id}` + `Retry-After`                               |
+| `204` | No Content           | `DELETE`                                                                                       |
+| `400` | Bad Request          | syntax: malformed JSON, missing required field, wrong type, bad parameter format              |
+| `401` | Unauthorized         | an unauthenticated write                                                                       |
+| `403` | Forbidden            | an authenticated caller without the required role                                             |
+| `409` | Conflict             | a reconcile whose `generation` is older than the stored one                                   |
+| `422` | Unprocessable Entity | well-formed but breaks a business rule: name collision, undefined property/tag, missing channel, empty/over-large reconcile |
+
+Never `200` with an error body — 2xx success, 4xx client fault, 5xx server fault.
+
+Other rules:
+
+- **Nouns in paths, plural and consistent.** `/channels`, `/properties`, `/tags` — never `/channel` or `/channels/process`.
+- **Errors use [RFC 9457 Problem Details](https://www.rfc-editor.org/rfc/rfc9457)** (`type`, `title`, `status` required; `detail`, `instance` optional) with content type `application/problem+json`. Use `"about:blank"` as `type` for generic HTTP errors.
+- **A command's per-item `results[]` is data, not an error body.** A `200` from `POST /channels:reconcile` means the command ran; the per-item statuses inside are its outcome. A malformed request returns `4xx` with a Problem Details body and no `results[]`. Per-item failures are never a `207 Multi-Status`.
+- **Request and response are separate schemas.** A write body holds only what the caller may set; the response holds the read projection.
+- **List and detail are separate representations.** A collection `GET` returns a lean summary; item `GET` and create/update responses return the full schema. Never use `null` to mean "omitted from the list"; reserve `null` for genuine optionality and list every nullable in `required`.
+- **Attachment is written only through the channel.** `/properties` and `/tags` define and read the vocabulary; membership is written only through `/channels/{channel_name}/properties` and `.../tags`.
+- **Two levels of nesting at most.** `/channels/{channel_name}/properties/{property_name}` is the limit.
+
+### Single-resource writes
+
+Plain methods handle one channel at a time, with an add-vs-replace split on the
+property and tag lists:
+
+| Method + Path                                   | Effect                                            |
+| ----------------------------------------------- | ------------------------------------------------- |
+| `POST /channels`                                | create one channel                                |
+| `PATCH /channels/{channel_name}`                | partial update                                    |
+| `DELETE /channels/{channel_name}`               | hard delete                                       |
+| `PUT /channels/{channel_name}/properties`       | replace the whole property list                   |
+| `PATCH /channels/{channel_name}/properties`     | add or update named properties, leaving the rest  |
+
+### Naming
+
+- **Lowercase kebab-case** for literal path segments.
+- **`snake_case`** for path params, query params, and JSON properties both directions (`channel_name`, `next_cursor`, `total_count`).
+- A filter or sort value that names a field uses the same spelling as the JSON property.
+- Sort is `sort=<field>:<direction>`, direction `asc` (default) or `desc` — never a boolean flag.
+- **`:camelCase`** for command segments. Standard hyphenated names for HTTP headers; never `snake_case` headers.
+
+### Operation IDs
+
+Every operation has a unique, stable `operationId` in `camelCase` (springdoc-openapi
+maps it to the generated controller method). Use `{verb}{Resource}` or
+`{verb}{Resource}{SubResource}` with a semantic verb, not the HTTP method
+(`createChannel`, not `postChannel`).
+
+| Verb                               | Use                                          |
+| ---------------------------------- | -------------------------------------------- |
+| `list`                             | collection `GET` returning many items        |
+| `get`                              | single-resource `GET`                        |
+| `query`                            | collection `GET` that is a search façade     |
+| `create`                           | `POST` that persists a new resource          |
+| `update`                           | `PATCH`, or `PUT` that replaces              |
+| `add` / `remove`                   | attach/detach a property or tag on a channel |
+| `delete`                           | `DELETE`                                      |
+| `reconcile` / `process` / `cancel` | colon-verb commands                          |
+
+| Endpoint                                             | operationId       |
+| ---------------------------------------------------- | ----------------- |
+| `GET /channels`                                      | `queryChannels`   |
+| `POST /channels`                                     | `createChannel`   |
+| `POST /channels:reconcile`                           | `reconcileChannels` |
+| `DELETE /channels/{channel_name}/tags/{tag_name}`    | `removeChannelTag` |
+| `DELETE /sources/{source_id}`                        | `deleteSource`    |
+| `POST /jobs/{job_id}:cancel`                         | `cancelJob`       |
+
+### Codegen
+
+An OAS-3.0 toolchain turns array-form nullables (`type: ["string", "null"]`) into
+`any`. Until a 3.1 generator is in use, write every nullable as `oneOf`:
+
+```yaml
+my_field:
+  oneOf:
+    - type: string
+    - type: "null"
+```
+
+## Authentication
+
+Access splits by method, not by path. Reads are open; writes need a role scoped
+to the resource; `admin` where noted.
+
+| Method class                    | Auth                | Role                                    |
+| ------------------------------- | ------------------- | --------------------------------------- |
+| `GET`                           | open (configurable) | —                                       |
+| `PUT` / `POST` / `PATCH` / `DELETE` | required        | `channel`, `property`, `tag`, or `admin` |
+
+- `401` for a missing/invalid credential; `403` for a valid one without the role.
+- **Role is the only gate.** v1 neither exposes nor checks an `owner`.
+- **Two transports: HTTP Basic and Bearer.** Bearer service tokens are recommended for machine producers like RecCeiver; Basic stays for compatibility. One scheme per request, never both. Roles come from group membership.
+
+## Creating Properties and Tags
+
+Properties and tags are created by `POST`ing a JSON body to the appropriate endpoint. You must create them first before they can be associated with channels. This is similar to how the current (1.9.6) RecCeiver does it.
+Attempting to create a property or tag that already exists will result in a `409 Conflict` error, hence its ok to send the request and ignore the explicit error.
+
+The only problem to note is that now properties have a `type` field that must be specified, if you configure multiple producers and one of them uses a different type for the same property name, you will need to update the type to match.
+Cleanup can only be done by deleting the property or tag and recreating it with the correct type.
+
+## Querying channels
+
+Channel search uses plain, named, documented parameters. Multiple parameters are
+**ANDed** — that conjunction of equality and glob filters is the whole
+first-surface algebra (no magic characters, no `GET`-with-body).
+
+| Parameter      | Filters on                    | Example                    |
+| -------------- | ----------------------------- | -------------------------- |
+| `name`         | channel name glob             | `name=SR*`                 |
+| `size` / `from` | offset paging                | `size=50&from=100`         |
+| `sort`         | `<field>:<direction>`         | `sort=name:asc`            |
+| `state`        | lifecycle                     | `state=active` / `inactive` / `any` |
+| `source_id`    | claiming source               | `source_id=recceiver-01:ioc-01` |
+| `tag`          | tag presence                  | `tag=active`               |
+| `prop_<name>`  | property value                | `prop_cell=2`              |
+| `cursor`       | opaque full traversal         | `?cursor=…`                |
+| `filter`       | RSQL (see below)              | `filter=…`                 |
+
+The `prop_` prefix keeps property selectors distinct from `name`, `tag`, and
+`state`. Negation, OR, and grouping are not in this surface — they move to RSQL.
+
+### Pagination
+
+Two contracts, both stateless at the API boundary:
+
+- **Offset** (`size` + `from`) — default for interactive discovery, bounded by the store's result window.
+- **Opaque `cursor`** — for full traversal (bulk export, "process all"), returned as `next_cursor`, passed back as `?cursor=…`.
+
+The cursor is opaque, so the engine's pagination primitive never reaches the
+contract. Query responses carry `{ total_count, channels[], next_cursor? }`.
+
+### RSQL
+
+When AND-of-filters is not enough, the richer layer is
+**[RSQL](https://github.com/jirutka/rsql-parser)** in a single `filter` parameter:
+
+```
+GET /channels?filter=(prop_cell==2,prop_cell==3);tag!=disabled
+```
+
+`;` is AND, `,` is OR, and parentheses group. It is a strict superset of the
+first-surface query capability. Four rules pin down semantics:
+
+1. **Selector namespacing.** Reserve `name`, `tag`, `state` for the channel's own fields; prefix property selectors with `prop_`.
+2. **Wildcards.** `*` and `?` in `==`/`!=` arguments are wildcards; matching is case-insensitive.
+3. **Absence vs. value.** `prop_cell!=*` is "property absent"; `prop_cell!=2` is "has `cell`, value ≠ 2".
+4. **Typed comparison.** A numeric property — integer or float — compares numerically and takes `<`/`>`/`<=`/`>=` (`prop_position>187`); a string property compares lexically and takes wildcards. The property's definition fixes which.
+
+### Aggregating property values
+
+**`GET /properties/{property_name}/values`** answers "what values does a property
+take, and how common is each?", scoped by the same parameters as `GET /channels`.
+So `GET /properties/ioc_id/values?name=SR:*` is "distinct `ioc_id`s among `SR:*`
+channels." The response is a values projection, sorted and paged over the values:
+
+```jsonc
+GET /properties/ioc_id/values
+{
+  "total_distinct": 42,
+  "values": [
+    { "value": "ioc01", "count": 130 },
+    { "value": "ioc02", "count": 118 }
+  ],
+  "next_cursor": "…"        // optional, same opaque-cursor contract as channels
+}
+```
+
+`sort` orders by `value` or `count` (`sort=count:desc` for the largest groups
+first); `size`/`from` and `cursor` page as for channels. The aggregation is
+opaque — the engine's own aggregation, exposing only values and
+counts.
+
+## Jobs (asynchronous operations)
+
+Some operations cannot finish in one request: a directory-wide processor sweep, or
+a bulk write of tens of thousands of channels. v1 models the work as a job
+resource the client submits and polls.
+
+| Method + Path                | operationId | Auth  | Purpose                        |
+| ---------------------------- | ----------- | ----- | ------------------------------ |
+| `GET /jobs/{job_id}`         | `getJob`    | open  | poll a job until terminal      |
+| `POST /jobs/{job_id}:cancel` | `cancelJob` | write | move a running job to canceled |
+
+Flow:
+
+1. The client submits the work. The server returns **`202 Accepted`** with a `Location: /jobs/{job_id}` header and a `Retry-After` hint. The body is the initial job.
+2. The client polls **`GET /jobs/{job_id}`** until the job is terminal.
+3. The job's `result` carries the outcome.
+
+```jsonc
+GET /jobs/{job_id}
+{
+  "job_id":     "…",
+  "kind":       "process_channels" | "reconcile_channels" | …,
+  "status":     "pending" | "running" | "succeeded" | "failed" | "canceled",
+  "created_at": "…",
+  "updated_at": "…",
+  "progress":   { "done": 8200, "total": 12000 },   // optional
+  "result":     { … },                              // present when terminal
+  "error":      { … }                               // RFC 9457, on infra failure
+}
+```
+
+| Field                    | Values                                                        | Notes                     |
+| ------------------------ | ------------------------------------------------------------- | ------------------------- |
+| `job_id`                 | string                                                        | identity                  |
+| `kind`                   | `process_channels` \| `reconcile_channels` \| …               |                           |
+| `status`                 | `pending` \| `running` \| `succeeded` \| `failed` \| `canceled` | terminal from `succeeded` on |
+| `created_at`/`updated_at` | timestamp                                                    |                           |
+| `progress`               | `{ done, total }`                                             | optional                  |
+| `result`                 | object                                                        | present when terminal     |
+| `error`                  | RFC 9457 Problem Details                                      | on infra failure          |
+
+- **Opting in.** `POST /channels:process` is always async — unbounded by directory size. `POST /channels:reconcile` is synchronous unless the client sends `Prefer: respond-async` ([RFC 7240](https://www.rfc-editor.org/rfc/rfc7240)), or the submission exceeds a configured size, in which case the server answers `202` with `Preference-Applied`.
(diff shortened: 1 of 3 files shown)


anderslindho:
Overall the v1 direction here is solid: plural nouns, correct method semantics, RFC 9457 errors, split request/response schemas, jobs for async work, RSQL replacing the old boolean-DSL magic chars. All good moves, agree with merging that as the baseline.

Where I think this needs another pass is the batch/bulk section, specifically re how it's meant to actually serve RecCeiver.

**batch vs reconcile**

`:batch`'s mixed post/put/patch/delete shape still needs the producer to know what already exists in CF (to bucket into post vs put) and what to delete (no query-scoped delete, no server-side reconcile, per the doc's own wording). That's the same read-before-write pattern we're trying to get RecCeiver away from. The doc even says it outright: "a producer that does not track what it holds builds the operations by diffing its desired set against a lean `GET /channels`...". If RecCeiver still has to GET-then-diff to build the batch, that's the same read, just done before submitting instead of after.

Proposal: add an actual `reconcile` op. Producer submits "this is my desired/observed set for scope X, generation N," CF upserts and tombstones anything in-scope not mentioned. That's the real one-way primitive. Once that exists I think the mixed-method `:batch` shape can go away entirely, reconcile covers what it was trying to do for RecCeiver, and I don't think anyone else needs mixed-method batch (open Q below).

Reconcile doesn't get to skip jobs/per-item results just because it's declarative. Still large-N (thousands of channels per IOC per our own README numbers), still partially-failable, and a failed item must not get tombstoned (it didn't disappear, it just failed validation), so reconcile needs per-item results plus sync-under-threshold/async-over-threshold, same machinery batch has today, just wrapped around a "desired state" body instead of an op-list.

**generic (non-reconcile) bulk, leaving this open**

Once reconcile absorbs RecCeiver: is there still a real need for generic bulk create/update/delete (admin tooling, migration scripts, whatever)? If yes, and it's small-N/interactive (our own actor table already has admin/curator as "occasional, deliberate"), I don't think that needs jobs or per-item results at all. Validate-then-write, atomic, single response and status. Simpler for callers than today's 200-with-embedded-422s. If there's actually a consumer doing large/failure-prone homogeneous bulk writes outside RecCeiver, that path needs per-item handling too, same as reconcile, just without tombstoning. Need to know if that consumer exists before deciding.

**200 vs 207 vs job, the doc contradicts itself here**

Hard rules say "never 200 with an error body." Async section says "never... a 207 Multi-Status." Sync `:batch`'s own example is a 200 with embedded per-item 422s. All three can't be true at once. If the intent is "`:batch` is a command, 200 = command ran, per-item statuses are data not errors," fine, but that needs to be written as an explicit carve-out to the hard rule, not left implicit.

**"mirrors ES `_bulk`", ground this in domain not backend**

Non-atomicity and the 200-status choice for batch are both justified by "follows Elasticsearch `_bulk`," but the doc is explicit elsewhere (opaque cursor) that engine behavior shouldn't leak into the contract. Same conclusion probably still holds, just ground it properly: large batches need forward progress, one bad item shouldn't sink 5k good ones, and no backend gives free cross-doc atomicity at this scale anyway. That reasoning would survive a hypothetical backend swap. The ES justification wouldn't.

**native concepts needed for reconcile**

For tombstoning to be safe I think we need, minimum:
- an opaque `source_id`/scope key, producer-supplied, CF doesn't interpret it (same idea as the opaque pagination cursor already in here)
- a generation/sequence guard per scope, so a delayed or reordered reconcile submission can't tombstone state a newer one already established
- first-class lifecycle fields on the channel (`state`, `last_seen_at`, `tombstoned_at`) instead of inferring lifecycle from arbitrary metadata, makes purge policy simpler to query, lets `GET /channels` default-exclude tombstoned channels, and gives an audit trail without extra work

Don't think we need IOC / RecCeiver / session to become their own CF resources though, scope just needs to be whatever unit appears/disappears together. RecCeiver owns getting that scoping right. CF just needs the opaque key plus generation guard.

**one more thing this forces us to decide now, not later: can one channel have multiple hosts?**

Separately from all of the above, PVs can be hosted by more than one IOC at once. Today CF pretends this can't happen, so when more than one recceiver/IOC reports the same name, whichever write lands last just silently clobbers the previous metadata. No visibility that it happened.

This lands directly on the `source_id`-scoped reconcile design above, not just future work. As sketched, if source A and source B both claim channel X, B's reconcile submission would silently overwrite A's metadata for X, and A's next reconcile would silently overwrite it back. Same bug we're trying to kill, just now inside the new primitive instead of the old one. So reconcile can't be specced without deciding: is an upsert a per-source claim on a channel, or a global clobber?

I think existence and metadata split cleanly here. Existence can aggregate now: channel exists if ≥1 source claims it, tombstone only when the last claim drops, that's just reference counting, solvable today. Metadata conflict is harder: two sources reporting different properties/tags for the same name doesn't aggregate cleanly, and I don't think we need to solve that now (whether we warn, reject, or just surface both isn't decided). But the identity question, keyed by name alone vs. some (name, ioc_id)-style composite, exact keying TBD, needs to at least be flagged as open now, because retrofitting single-owner -> multi-claim later is a lot more expensive than leaving room for it in the reconcile primitive from the start.

jacomago:
> **batch vs reconcile**
> 
I can swap to reconcile.

> **native concepts needed for reconcile**
> 
> For tombstoning to be safe I think we need, minimum:
> 

I still don't understand tombstoning, is this a feature inside of channelfinder? Or is this a client is pushing some extra metadata?

Can you give an example of using just the normal update a single channel example?

> 
> **one more thing this forces us to decide now, not later: can one channel have multiple hosts?**
> 

urgl, I don't particularly want to tackle this as it requires doing an elastic migration. The channel name is the key in the channels 'table'. Adding a source_id (which I think would have to be `recceiver:ioc_id`) Would mean adding a migration to elastic which we've never done before. The support for migrations also kind of ... is not nearly as good as flyway is for postgres.

Which completely changes the scope of this PR. Adding support for a source_id, which I think is pretty easily added on top instead, we modify the Channel Response and Request object with an optional field source_id. This would not break a v1 api. I think it does not need to be added now for this reason.

dev-docs/api/conventions.md
```diff
@@ -0,0 +1,397 @@
...
+not the current API, which predates them; migrations.md covers the gap and the
+rationale.
+
+The domain has three resources:
+
+- **channels** — a `name` (unique identifier), an `owner`, a set of properties, and a set of tags.
+- **properties** — a name-value definition (`name`, `owner`) attachable to channels with a per-channel `value`.
+- **tags** — a name-only label (`name`, `owner`) attachable to channels.
+
+Two operational resources sit alongside the domain — addressable resources that
```
simon-ess: Typo: shouldn't this be `domain-addressable`?

dev-docs/api/conventions.md
```diff
@@ -0,0 +1,397 @@
...
+- **List vs detail representations** — a collection `GET` returns a lean summary schema; an item `GET` and create/update responses return the full schema. Never use `null` to mean "omitted in the list projection" — split the schema instead. Reserve `null` for genuine domain optionality.
+- **Request and response are separate schemas.** A write body holds only what the caller may set; the response holds the read projection. Never reuse one schema for both, and never put a field on a write body only to ignore it (e.g. a definition's `channels[]` on a property write).
+- **Attachment is written only through the channel.** `/properties` and `/tags` define and read the vocabulary; a channel's membership is written only through `/channels/{channel_name}/properties` and `/channels/{channel_name}/tags`. No second path mutates membership from the definition side.
+
+## Authentication & authorization
+
+Access is split by HTTP method, not by path:
+
+- **Reads (`GET`) are unauthenticated.**
```
simon-ess: Should this perhaps be optional? Given that we seem to be moving towards all services requiring authentication here at ESS, should we not consider this configurable?

dev-docs/api/conventions.md
```diff
@@ -0,0 +1,397 @@
...
+- **Writes (`PUT`/`POST`/`PATCH`/`DELETE`) require authentication** and a role scoped to the resource type — `channel`, `property`, `tag`, or `admin`. Roles derive from group membership.
+- `401` for a missing/invalid credential; `403` for a valid credential without the required role.
+- **Role is the only authorization gate.** An `owner` is recorded for provenance but not enforced per call — a caller with the right role may write a resource owned by another.
+- **Two transports: HTTP Basic and Bearer tokens.** Bearer application/service tokens are recommended for machine producers (e.g. RecCeiver); Basic is kept for compatibility. A request carries exactly one scheme — never both in one `Authorization` header.
+
+## Naming
+
+- **Lowercase kebab-case** for literal path segments — `/channels`, `/channels/{channel_name}/properties`.
```
simon-ess: This is a bad example.  There are no kebabs here.

dev-docs/api/conventions.md
```diff
@@ -0,0 +1,397 @@
...
+- `PATCH /processors/{processor_name}` → `updateProcessor`
+- `GET /jobs/{job_id}` → `getJob`
+- `POST /jobs/{job_id}:cancel` → `cancelJob`
+
+## Querying channels
+
+Channel search uses plain, named, documented query parameters:
+
+- Name glob, page size, and sort: `GET /channels?name=SR*&size=50&sort=channel_name:asc`.
+- Tag presence and property values are named parameters too — `tag=active`, `cell=2`.
```
simon-ess: There is some ambiguity here: What if I have a property named `tag`?

dev-docs/api/conventions.md
```diff
@@ -0,0 +1,397 @@
...
+reconcile that must drop disappeared metadata uses `put` with the full desired
+state.
+
+- **Per-item, not transactional.** Each item is applied and reported independently; a `200`/`job.succeeded` means the batch ran, not that every item passed. The outcome is one `results[]`, keyed by channel `name` (present on every item, including `delete`, so no correlation token is needed): each entry carries its `status`, a failure its RFC 9457 `error`. A sync batch returns it inline, an async batch in `job.result`.
+- **Defined order.** Creates (`post`), then replaces (`put`), then merges (`patch`), then deletes (`delete`); document order within a method. No cross-item referencing, and a name appearing twice is a malformed document (`400`).
+- **Sync by default, async on request.** The batch runs inline unless the client sends `Prefer: respond-async` (or it exceeds the server's size threshold), in which case it returns a job.
+
+```jsonc
+{ "results": [
+  { "name": "SR:C01", "method": "post",   "status": 201 },
```
simon-ess: If we go this direction (I share @anderslindho 's concerns) shouldn't this return the entire object instead of just a 201?

dev-docs/api/conventions.md
```diff
@@ -0,0 +1,397 @@
...
+
+```jsonc
+POST /channels:batch
+{
+  "post":   [ { "name": "SR:C01", "properties": { "cell": "1" }, "tags": ["active"] } ],
+  "put":    [ { "name": "SR:C02", "properties": { "cell": "2" }, "tags": [] } ],
+  "patch":  [ { "name": "SR:C03", "properties": { "cell": "3" } } ],
+  "delete": [ "SR:C99" ]
+}
+```
```
simon-ess, line 295: What is the use case for this instead of allowing
```jsonc
POST /channels:batch
{
   [ { "name": "SR:C01", "properties": { "cell": "1" }, "tags": ["active"] } ]
}
```
vs
```jsonc
PUT /channels:batch
{
  [ { "name": "SR:C02", "properties": { "cell": "2" }, "tags": [] } ]
}
```
etc.? Why do we need to create, update, update, and delete all at once?

dev-docs/api/conventions.md
```diff
@@ -0,0 +1,397 @@
...
+  "updated_at": "…",
+  "progress":   { "done": 8200, "total": 12000 },   // optional
+  "result":     { … },                              // present when terminal
+  "error":      { … }                               // RFC 9457, on infra failure
+}
+```
+
+A job is a noun with a lifecycle: async work is server state, so it is an
+addressable resource, not a hidden session id. Jobs are retained for a bounded
+window, then reaped.
```
simon-ess: Do we need to reap these? I think of awx where the jobs are there in perpetuity.

That said, maybe we don't care to revisit job logs years after the fact 🙃

dev-docs/api/conventions.md
```diff
@@ -0,0 +1,397 @@
...
+**`job.succeeded` ≠ every item succeeded.** A batch's outcome is per-item — some
+channels persist, some fail a business rule. That per-item report is one schema,
+two delivery modes: a synchronous batch returns the `results[]` array inline; an
+async batch puts the identical `results[]` in `job.result`. Each failed item
+embeds its own RFC 9457 problem detail. A job reaching `succeeded` means the
+operation ran to completion, not that every item passed — failures live in the
+results, never collapsed into a `200`-with-error-body or a `207 Multi-Status`.
+
+**Idempotent submission.** A submission may carry an **`Idempotency-Key`** header;
+a repeat with the same key returns the same job instead of starting a second.
```
simon-ess: Depends how long you keep the jobs for!

simon-ess:
> I still don't understand tombstoning, is this a feature inside of channelfinder? Or is this a client is pushing some extra metadata?

From what I read (e.g. https://en.wikipedia.org/wiki/Tombstone_(data_store)) it seems not obviously relevant in our case as we do not have a distributed database to deal with, but perhaps it is the correct approach regardless?

dev-docs/api/conventions.md
jacomago: Nice catch.

dev-docs/api/conventions.md
jacomago: Well, my thought was that recceiver would be the one deleting channels, but this is really the crux between 'resolve' vs 'batch'. Does the recceiver calculate the sync or does channelfinder.

dev-docs/api/conventions.md
jacomago: Process would run after every single ioc restart. So thats a lot of jobs.

simon-ess: Eh... it's not like we will run out of integers. And looking at other utilties, we have typically one or more jobs per IOC start, stop, deploy, undeploy, which is of the same order....

anderslindho:
> Well, my thought was that recceiver would be the one deleting channels, but this is really the crux between 'resolve' [sic] vs 'batch'. Does the recceiver calculate the sync or does channelfinder.

ChannelFinder - recCeiver always just pushes what it has. Else it needs state/db. Good that you're up for swapping to reconcile. @simon-ess's "why do we need to create, update, update, and delete all at once?" on the batch example is the same question from the other end I think.

> I still don't understand tombstoning, is this a feature inside of channelfinder? Or is this a client is pushing some extra metadata?

> Can you give an example of using just the normal update a single channel example?

> (SR) _From what I read (e.g. https://en.wikipedia.org/wiki/Tombstone_(data_store)) it seems not obviously relevant in our case as we do not have a distributed database to deal with, but perhaps it is the correct approach regardless?_

TBH I am not 100% sure if it is the correct term, but I think it covers it. What I mean by it is that CF alone is responsible for changing channels from active to inactive. So my _intention_ is basically soft delete + retention/purge policy, and the important part is who owns that (CF).

To answer the either/or directly: today it is a client pushing extra metadata. That's `pvStatus`. Client decides a channel went away, writes that decision into CF as a property. That's the pattern I want gone.

**Example**

Say `SOME:AI` stops being reported.

Today: recCeiver reads CF, works out that `SOME:AI` is no longer in its set, and writes the deactivation itself (`pvStatus` -> `Inactive`, or a delete). CF is a dumb store, the client holds the logic.

Proposed: recCeiver sends its current set for its scope, `SOME:AI` isn't in it, CF notices it's in-scope-but-absent and does the equivalent of `PATCH /channels/SOME:AI {"state": "inactive"}` on its own. Record stays, `GET /channels` excludes it by default, it comes back active on the next upsert that mentions it. recCeiver never reads and never computes a diff.

> **one more thing this forces us to decide now, not later: can one channel have multiple hosts?**

> urgl, I don't particularly want to tackle this as it requires doing an elastic migration. [...]

> Which completely changes the scope of this PR. Adding support for a source_id, which I think is pretty easily added on top instead, we modify the Channel Response and Request object with an optional field source_id. This would not break a v1 api. I think it does not need to be added now for this reason.

I think we should separate API design from business logic design. What I think is important is that the API already now accommodates for this design. How exactly that is implemented and/or how dbs are migrated to support it is a separate issue.

Not asking for the migration, and definitely not in this PR. But "add it on top later" only holds for the field, not for the semantics. Schema-wise you're right, optional `source_id` is additive. Semantically no: whether an upsert is a per-source claim or a global clobber is observable behavior clients build on. Change it later and the same request does something different.

And it's not really future work, it's load-bearing for reconcile being correct. With a single `source_id`, a channel hosted by two IOCs gets deactivated by whichever one drops it first, even though the other still has it. Live PVs flapping in and out of `GET /channels`. The old failure was two writes clobbering each other; the new one is CF soft-deleting things on its own bad inference. Worse, because now it's our logic doing it.

Doesn't mean solving it now. But reconcile's rule needs to be "deactivate when nothing claims it" rather than "deactivate when the recorded source dropped it", and v1 can implement that with one claim per channel as a documented limitation. Free to write down now, expensive once clients have built on one-source-per-channel being the actual rule.

shroffk:
> > one more thing this forces us to decide now, not later: can one channel have multiple hosts?
> 
> Separately from all of the above, PVs can be hosted by more than one IOC at once. Today CF pretends this can't happen, so when more than one recceiver/IOC reports the same name, whichever write lands last just silently clobbers the previous metadata. No visibility that it happened.

During the meeting we agreed that I will be adding a "preprocessor" SPI similar to the "processor" one that exists.

So each site can add their own validation for channels... one of these can be a validator that detects if existing/active channels are there in channelfinder which have different hosts, it can then log or notify the appropriate entities to resolve the issue.

I don't think we should allow multiple properties with the same name.

tynanford:
Overall I really like this. RSQL queries would be great.

I'm also thinking more about the duplicate PV names. I agree with @anderslindho that it is good to think about even if we don't address/support it fully in v1. I'm not saying this is practical but if PV name is the key and a PV can have multiple `iocid`, `hostname`, etc., then a channel in CF could be represented like this:

```json
{
    "name": "SR12C:SQSF1:Setpoint",
    "properties": {
      "Acc": "SR",
      "Family": "SQSF",
      "Field": "Setpoint",
      "Sector": "12",
      "Device": "1",
      "Position": "187.277343",
      "hostName": "ioc-host-a.example.local",
      "iocName": "skewquads2",
      "iocid": "192.168.20.11:45334",
      "iocIP": "192.168.20.11",
      "time": "2026-07-15 09:12:00.123456",
      "caPort": "45502",
      "recordType": "ao",
      "recordDesc": "Current setpoint",
      "archive": "Slow",
      "Engineer": "jdoe",
      "Location": "SoftIOC"
    },
    "pvStatus": "Active",
    "conflict": {
      "duplicateActiveSources": true,
      "activeSourceIds": [
        "recceiver-1.example.local:192.168.20.11:45334",
        "recceiver-2.example.local:192.168.20.12:52110"
      ],
      "note": "flattened properties above reflect one of these sources arbitrarily; do not treat as authoritative until resolved"
    },
    "tags": ["MML", "PS"],
    "sources": [
      {
        "source_id": "recceiver-1.example.local:192.168.20.11:45334",
        "pvStatus": "Active",
        "last_seen": "2026-07-15T09:12:00Z",
        "properties": {
          "recceiverID": "recceiver-1.example.local",
          "iocName": "skewquads2",
          "hostName": "ioc-host-a.example.local",
          "iocid": "192.168.20.11:45334",
          "iocIP": "192.168.20.11",
          "time": "2026-07-15 09:12:00.123456",
          "caPort": "45502",
          "recordType": "ao",
          "recordDesc": "Current setpoint",
          "archive": "Slow",
          "Engineer": "jdoe",
          "Location": "SoftIOC"
        }
      },
      {
        "source_id": "recceiver-2.example.local:192.168.20.12:52110",
        "pvStatus": "Active",
        "last_seen": "2026-07-15T09:10:45Z",
        "properties": {
          "recceiverID": "recceiver-2.example.local",
          "iocName": "skewquads2-test",
          "hostName": "ioc-host-b.example.local",
          "iocid": "192.168.20.12:52110",
          "iocIP": "192.168.20.12",
          "time": "2026-07-15 09:10:45.987654",
          "caPort": "45519",
          "recordType": "ao",
          "recordDesc": "Current setpoint (test IOC)",
          "archive": "Slow",
          "Engineer": "jdoe",
          "Location": "SoftIOC"
        }
      },
      {
        "source_id": "recceiver-1.example.local:192.168.20.13:38821",
        "pvStatus": "Inactive",
        "last_seen": "2026-06-20T14:22:03Z",
        "properties": {
          "recceiverID": "recceiver-1.example.local",
          "iocName": "skewquads2",
          "hostName": "ioc-host-old.example.local",
          "iocid": "192.168.20.13:38821",
          "iocIP": "192.168.20.13",
          "time": "2026-06-20 14:22:03.456789",
          "caPort": "45506",
          "recordType": "ao",
          "recordDesc": "Current setpoint",
          "archive": "Slow",
          "Engineer": "jdoe",
          "Location": "SoftIOC"
        }
      }
    ]
  }
```

Each source in this case is an IOC connection to a specific recceiver instance and includes all the properties that IOC reported. The top level properties is the flattened view for normal clients and includes properties manually added by non-recceiver clients. The top level pvStatus is an OR of all the source pvStatus fields and there is a conflict flag in case there are multiple active sources.

There are some problems with this idea (including an IOC restart adding a new source every time) but throwing it out there to brainstorm.

tynanford:
oh that json blob reminded me of an old complaint about CF here: all the property values are strings so things like position in the example above are hard to sort on unless you make all positions the same number of characters. We added leading zeros in those cases. Not sure if it can be addressed or makes sense but did come up here

jacomago:
Changes:

- Tried to simplify the documents
- Added sources, a more minimal version of what @tynanford added but with a lot of the same features
  - I went for a default of source_id: 'recceiver_id:ioc_name' as 'iocid' has problem that restarts gets new ports and hence new ioc_ids. But I also put its up to the user of the api to choose the source_id (i.e. the recceiver)
  - didn't track old property values, not sure how useful that is from just knowing there are multiple sources and then people can fix
- Swapped to reconcile rather than batch
  - Added generation for when two reconciles conflict
- Added design of a timeout on channels existing after inactive, same with sources
- dropped the owner
- basically wrote how lifecycle of a channel could be handled by ChannelFinder itself, tried to add a comment about how that could still work with v0 api.
- Add that property value can be float, integer or string, which i was thinking of adding before anyway.

anderslindho (dismissed):
Nice, this took on basically everything that I hoped for. Few things still open.

Biggest is `source_id` granularity. Mechanism is fine, but on the recommended `recceiver_id:ioc_name`: "IOC-name" is an ESS-ism, and, although it can be an example, it should not be the canonical form. And it might be useful to state the actual rule, which is that scope has to be the smallest thing that appears/disappears together. Same PV under multiple IOCs on one recCeiver only works if you scope per-IOC - then X has a claim from ioc-A and ioc-B and stays active till both drop. Scope per-recCeiver and one IOC dropping deactivates PVs the other still serves. So IMO state the granularity rule, and say same-PV-under-multiple-scopes is expected multi-claim, not an error.

Batch being gone is good, but it leaves no generic bulk primitive at all - bulk just means reconcile now. Anything non-recCeiver wanting high-volume writes (admin, migration) is stuck looping singles or faking a source_id. Completely fine for me, but _maybe_ worth stating explicitly - your call.

Reconcile items reference properties/tags and an undefined one 422s per-item, so a recCeiver pushing a new property has to define it first or those items fail. That pre-provision step cuts against one-way. Should reconcile create missing definitions, or do producers pre-provision and we document it?

`conflict:true` is a good call. But: you can see two sources disagree but not what the losing value was (probably ok since we dropped per-source props but might warrant call-out), and doesn't say when conflict clears.

Minor: `source_id` optional with host fallback reintroduces the host instability we're avoiding, for whoever doesn't set it. Might be worth making required.

None of this blocks merge for me.

jacomago:
> Nice, this took on basically everything that I hoped for. Few things still open.
> 
> Biggest is `source_id` granularity. Mechanism is fine, but on the recommended `recceiver_id:ioc_name`: "IOC-name" is an ESS-ism, and, although it can be an example, it should not be the canonical form. And it might be useful to state the actual rule, which is that scope has to be the smallest thing that appears/disappears together. Same PV under multiple IOCs on one recCeiver only works if you scope per-IOC - then X has a claim from ioc-A and ioc-B and stays active till both drop. Scope per-recCeiver and one IOC dropping deactivates PVs the other still serves. So IMO state the granularity rule, and say same-PV-under-multiple-scopes is expected multi-claim, not an error.
> 

That's why I made it recommended, its up to the producer to decide but something per-IOC is best. And something that survives reconnects of IOCs, which iocid does not. But I'll try make language clearer about what I mean by 'recommended'.

> Reconcile items reference properties/tags and an undefined one 422s per-item, so a recCeiver pushing a new property has to define it first or those items fail. That pre-provision step cuts against one-way. Should reconcile create missing definitions, or do producers pre-provision and we document it?
> 

I will add produces pre-provision. This is how recceiver works now, and since property list is small I don't see a big problem with checking properties exist or just pushing the properties at start.

> `conflict:true` is a good call. But: you can see two sources disagree but not what the losing value was (probably ok since we dropped per-source props but might warrant call-out), and doesn't say when conflict clears.
> 

I might just drop this, multiple sources always suggests a conflict of some kind. Could just be latest time seen.

> Minor: `source_id` optional with host fallback reintroduces the host instability we're avoiding, for whoever doesn't set it. Might be worth making required.
> 
Updated the fallback to be based on iocname and iocid property + the host

> None of this blocks merge for me.

tynanford:
At our site we sometimes struggle with the duplicate IOCs being started on the same host and `recceiver_id:ioc_name` won't catch that. Using `iocid` has the problem of the port but what about an optional feature that any inactive sources are removed when a new active source appears? Maybe one issue would be a race condition between an IOC restarting and recceiver/CF changing the old source_id to inactive. Then you have 2 active sources for a bit, which might be an issue for how to clean that up later.

You also then lose the ability to see the history of a PV moving from IOC-A to IOC-B. Hmm.

I'm not sure we would want the purge timeout here but I assume that can be disabled/set to 0. If a PV goes inactive because it was retired, we want that PV to stay in CF for a while in case people search for it.

simon-ess:
> At our site we sometimes struggle with the duplicate IOCs being started on the same host and `recceiver_id:ioc_name` won't catch that. Using `iocid` has the problem of the port but what about an optional feature that any inactive sources are removed when a new active source appears? Maybe one issue would be a race condition between an IOC restarting and recceiver/CF changing the old source_id to inactive. Then you have 2 active sources for a bit, which might be an issue for how to clean that up later.

I'm not sure I follow here. We have hosts with multiple IOCs, so how would you know which one to remove when a new one appears?

Would it be possible to have the reccasters send an optional `iocid` parameter which is used, can be site-specific, and defaults to the old iocid if not defined or if a collision occurs?

dev-docs/api/conventions.md
```diff
@@ -0,0 +1,536 @@
...
+- **A command's per-item `results[]` is data, not an error body.** A `200` from `POST /channels:reconcile` means the command ran; the per-item statuses inside are its outcome. A malformed request returns `4xx` with a Problem Details body and no `results[]`. Per-item failures are never a `207 Multi-Status`.
+- **Request and response are separate schemas.** A write body holds only what the caller may set; the response holds the read projection.
+- **List and detail are separate representations.** A collection `GET` returns a lean summary; item `GET` and create/update responses return the full schema. Never use `null` to mean "omitted from the list"; reserve `null` for genuine optionality and list every nullable in `required`.
+- **Attachment is written only through the channel.** `/properties` and `/tags` define and read the vocabulary; membership is written only through `/channels/{channel_name}/properties` and `.../tags`.
+- **Two levels of nesting at most.** `/channels/{channel_name}/properties/{property_name}` is the floor.
```
simon-ess: Shouldn't that be `ceiling`, not `floor`?

dev-docs/api/conventions.md
```diff
@@ -0,0 +1,536 @@
...
+}
+```
+
+- `last_seen_at` is per source, not per channel. "When was this channel last seen?" — an active channel takes the most recent `last_seen_at` among the sources that claim it active; an inactive one takes its most recent link `updated_at`.
+- A source's **`state`** is derived from its links: `active` while it holds at least one active link, `inactive` once all its links are inactive.
+
+| Method + Path                 | operationId   | Auth    | Purpose                                                     |
+| ----------------------------- | ------------- | ------- | ---------------------------------------------------------- |
+| `GET /sources`                | `listSources` | open    | source records; default-excludes inactive (`state=inactive`/`any` to include) |
+| `GET /sources/{source_id}`    | `getSource`   | open    | one record                                                 |
```
simon-ess: I don't understand `Purpose: one record`. This should get the info of a single source (an IOC/recceiver combination) but what  do you mean by "one record"?

dev-docs/api/conventions.md
```diff
@@ -0,0 +1,536 @@
...
+
+```
+generation = max(now_ms, last_current_generation + 1)
+```
+
+seeding `last_current_generation` from the most recent `409`'s
+`current_generation`. This turns "fast clock wins forever" into "fast clock wins
+this round; the loser re-syncs on its next snapshot."
+
+Two redundant producers sharing one `source_id` is a deliberate choice, not a
```
simon-ess, line 462: When  will two producers share the same source ID? Does that mean that the recceivers are configured with the same ID and they both find an IOC with the same name/id?

jacomago: Yes exactly. Two recceivers on same subnet, you might want them as a redundant pair. One gets restarted while the other stays up, the source_id stays the same and conflicts are handled by the generation param.

dev-docs/api/conventions.md
jacomago: Just not multiple like /sources.

dev-docs/api/conventions.md
jacomago: limit 😄

dev-docs/api/conventions.md
simon-ess: I think it's odd that it's called `record` though. A record can be many things, but in the context of channelfinder it's hard not to think of it as an IOC database record.

jacomago: swapped to source info

tynanford:
> I'm not sure I follow here. We have hosts with multiple IOCs, so how would you know which one to remove when a new one appears?

The case I am thinking of is PV `ABC` exists in IOC `XYZ` which connects to `recceiver-a`. If XYZ IOC gets started twice on the same host/subnet, then each IOC instance will have the same source ID: `recceiver-a:XYZ` and CF won't catch the fact that there are actually two instances of PV `ABC`. Whereas using `recceiverid:iocid` would correctly show there are two sources here.

With iocid there is the issue of an IOC restart creating a new `source_id`. I think it is more honest though since iocid uniquely identifies the IOC/recceiver connection unlike just iocname/recceiverid

My question is can we make this workable, i.e. what about if a new active source appears, the inactive sources get purged automatically? Then IOC restarts don't continuously increase the number of sources.

jacomago:
> My question is can we make this workable, i.e. what about if a new active source appears, the inactive sources get purged automatically? Then IOC restarts don't continuously increase the number of sources.

Yeah thats a good idea. I'll update the section. Essentially what your saying is a source retires when all its links are inactive, and then after 'source-timeout-config' it gets deleted. 

I think now it says more that when all channels its linked to are inactive.

anderslindho:
Nuance on IOC-name: at ESS it's not just a name, it comes from our inventory/deployment system, so it's an actual uniqueness guarantee. That's what makes it usable as an identifier for us. But that's an ESS thing - other sites (Tynan's included) don't have an IOC-name at all, or nothing backing it, so it can't carry that weight for them.

I also think two separate problems are getting mixed together: identity/uniqueness (what identifies a source, and how you spot duplicate IOC instances, with no inventory to lean on) and reaping (cleaning up inactive/dead sources). Tynan's iocid + purge idea solves the first by way of the second, and the timeout answers the second but not the first. They probably want different answers - worth keeping apart.

And I'm a bit on the fence about auto-deleting sources. Even the timeout version eventually erases the IOC-A -> IOC-B history Tynan said he wanted to keep, and it's purely about reaping - doesn't touch the duplicate/identity problem at all.

tynanford:
> Nuance on IOC-name: at ESS it's not just a name, it comes from our inventory/deployment system, so it's an actual uniqueness guarantee. That's what makes it usable as an identifier for us. But that's an ESS thing - other sites (Tynan's included) don't have an IOC-name at all, or nothing backing it, so it can't carry that weight for them.

Ah I see. Yes our IOC names are just the variable in the st.cmd file so there can be duplicates.

> And I'm a bit on the fence about auto-deleting sources. Even the timeout version eventually erases the IOC-A -> IOC-B history Tynan said he wanted to keep, and it's purely about reaping - doesn't touch the duplicate/identity problem at all.

The history would be nice but that is lower on the priority IMO. Most important with the sources is to tell if there are multiple active sources. I guess this reaping/cleanup is more an implementation detail? I would push for it to be an option at least to auto-purge inactive sources when a new active source appears. The goal would be something like this:

Single live PV: 1 source (active)
Duplicate PVs: Multiple sources (all active)
Retired PV: 1 source (inactive)

